PR #65, opened 2022-01-18
Ret278 refactor venue addresses

Angela-LW:
https://tools.hmcts.net/jira/browse/RET-278

### Change description ###
Replaced old excel venue addresses list with .yml version

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/VenueAddressReaderService.java
```diff
@@ -0,0 +1,78 @@
...
+                return hearingType.getHearingGlasgow().getSelectedLabel();
+            case ABERDEEN_OFFICE:
+                return hearingType.getHearingAberdeen().getSelectedLabel();
+            case DUNDEE_OFFICE:
+                return hearingType.getHearingDundee().getSelectedLabel();
+            case EDINBURGH_OFFICE:
+                return hearingType.getHearingEdinburgh().getSelectedLabel();
+            default:
+                log.error("No {} venue found", venue);
+                return null;
```
jthmcts, line 75: I think it would be better to throw an exception rather than return null

hsjhita (approved):

jthmcts (approved):

PR #89, opened 2022-02-09
Ret 887 fix hearings to judgment report

Angela-LW:
https://tools.hmcts.net/jira/browse/RET-887

### Change description ###
- Fixed Hearings to Judgments report so that it works with new CCD configs
- Refactored report data generation classes and moved report specific code from common package to this one
- Fixed related unit tests

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/reports/Constants.java
```diff
@@ -2,15 +2,29 @@
```
gauravchawla73463, line 2: I was wondering if it would be a good idea to have constants class for individual report to be clear that total_within_4weeks is for hearings to judgment report for example

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/reports/hearingstojudgments/HearingsToJudgmentsCcdDataSource.java
```diff
@@ -2,25 +2,35 @@
...
         try {
-            var query = HearingsToJudgmentsElasticSearchQuery.create(listingDateFrom, listingDateTo);
-            return ccdClient.hearingsToJudgementsSearch(authToken, caseTypeId, query);
+            var query = HearingsToJudgmentsElasticSearchQuery.create(managingOffice, listingDateFrom, listingDateTo);
+            var submitEvents = new ArrayList<HearingsToJudgmentsSubmitEvent>();
+            var searchResult = ccdClient.runElasticSearch(authToken, caseTypeId, query,
+                    HearingsToJudgmentsSearchResult.class);
+
+            if (searchResult != null && CollectionUtils.isNotEmpty(searchResult.getCases())) {
+                submitEvents.addAll(searchResult.getCases());
```
gauravchawla73463, line 30: May be just return searchResult.getCases() without having to declare submitEvents?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/reports/Constants.java
Angela-LW: I've moved the not shared constants into their respective report data classes where they're used

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/reports/hearingstojudgments/HearingsToJudgmentsCcdDataSource.java
Angela-LW: refactored to just return the case list rather than setting another variable

PR #231, opened 2022-05-03
RET-862

hsjhita:

https://tools.hmcts.net/jira/browse/RET-862

### Change description ###

Validation for adding singles to a multiple from a different office

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/excel/MultipleCreationMidEventValidationService.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/excel/MultipleCreationMidEventValidationService.java
@@ -18,6 +18,7 @@ import static uk.gov.hmcts.ecm.common.model.helper.Constants.ACCEPTED_STATE;
 import static uk.gov.hmcts.ecm.common.model.helper.Constants.ET1_ONLINE_CASE_SOURCE;
 import static uk.gov.hmcts.ecm.common.model.helper.Constants.MANUALLY_CREATED_POSITION;
 import static uk.gov.hmcts.ecm.common.model.helper.Constants.MIGRATION_CASE_SOURCE;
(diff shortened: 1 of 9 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/excel/MultipleCreationMidEventValidationService.java
```diff
@@ -164,6 +170,14 @@ private void validateSingleCasesState(List<SubmitEvent> submitEvents, List<Strin
 
             }
 
+            if (!isScotland && !isNullOrEmpty(submitEvent.getCaseData().getManagingOffice())
```
jthmcts, line 173: Do we need this granularity in the validation of Scottish cases? Do we care about managing office for Scotland?

hsjhita: Based off the current implementation, no. In Scotland, cases within a multiple can be part of different offices. The work I have done get's it to the same state as ECM. From what you are saying, Glasgow singles should only be added to Glasgow multiples but this has never been the case on ECM

hsjhita: I am getting this checked though whether it should be implemented

jthmcts: Yeah sorry I didn't read the code carefully enough. Yes it should do the same as ECM does now. You could argue it would be a nicer API if the parameter was caseTypeId rather than isScotland then you can put the logic to determine if it is Scotland in that method rather than ask clients to work it out, but I'm being uber picky there

jthmcts (approved):

marcjudge (approved):

PR #245, opened 2022-05-09
Update how ethos case references are generated

jthmcts:
https://tools.hmcts.net/jira/browse/RET-822

### Change description ###
Change how ethos case refs are generated. For England/Wales refs start at 6000001 and Scotland 8000001. References increment by 1 and are reset for a new year. This applies to both singles and multiples.

The postgres functions have also been reworked. There are now separate functions for England/Wales and Scotland.
The parameter for num_cases has been removed as this is redundant.

flyway integration has been changed so that migrations are applied when the application boots up rather than in the Jenkins pipeline. This means that the initial development action of creating a database no longer uses the gradle installlocaldb. Instead a developer should just call ./bin/init-db.sh

This is a breaking change as I have amended the existing flyway migration scripts rather than create new ones. This decision was taken as the application is still pre-prod and it is a significantly neater and simpler solution.

```
[x] Yes
[ ] No
```


src/main/resources/application-dev.properties
```diff
@@ -1,5 +1,7 @@
 spring.datasource.url=jdbc:postgresql://localhost:5050/et_cos
 
+spring.flyway.locations=classpath:db/migration,classpath:db/dev
```
hsjhita, line 3: isn't this also needed in application.properties as well?

jthmcts: Migration scripts are read from db/migration by default.

https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html#howto.data-initialization.migration-tool.flyway

hsjhita (approved):

PR #259, opened 2022-05-11
Ret 1778

robin-ieong:
[RET-1778](https://tools.hmcts.net/jira/browse/RET-1778)

Enable an admin user to be able to add a new judge to the ECM reference data.

The functionality should be available from the ECM Admin case type.

The ECM Admin case type should include a new event/step called Add judge.

The event should input the tribunal office, judge code, judge name, judge employment status.

tribunal office = droplist of all offices in England/Wales plus single entry for Scotland.

code = textbox

name = textbox

employment status = dropdown list with entries Fee Paid, Salaried, Unknown

All fields are mandatory.

The judge code and name should not already exist for the tribunal office.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/domain/repository/JudgeRepository.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/domain/repository/JudgeRepository.java
@@ -10,4 +10,6 @@ import java.util.List;
 @Repository
 public interface JudgeRepository extends JpaRepository<Judge, Integer> {
     List<Judge> findByTribunalOffice(TribunalOffice tribunalOffice);
+
+    boolean existsByCodeOrName(String code, String name);
 }
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/admin/AddJudgeController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/admin/AddJudgeController.java
@@ -1,0 +1,53 @@
+package uk.gov.hmcts.ethos.replacement.docmosis.controllers.admin;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import lombok.RequiredArgsConstructor;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.PostMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestHeader;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+import uk.gov.hmcts.ethos.replacement.docmosis.domain.admin.AdminData;
+import uk.gov.hmcts.ethos.replacement.docmosis.domain.admin.CCDCallbackResponse;
+import uk.gov.hmcts.ethos.replacement.docmosis.domain.admin.CCDRequest;
+import uk.gov.hmcts.ethos.replacement.docmosis.service.VerifyTokenService;
+import uk.gov.hmcts.ethos.replacement.docmosis.service.admin.AddJudgeService;
+
+import static org.springframework.http.MediaType.APPLICATION_JSON_VALUE;
+
+@RestController
+@RequestMapping("/admin")
+@RequiredArgsConstructor
+public class AddJudgeController {
+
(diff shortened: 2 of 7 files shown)


src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/admin/JudgeServiceTest.java
```diff
@@ -0,0 +1,60 @@
...
+    void setup() {
+        judgeRepository = mock(JudgeRepository.class);
+    }
+
+    @Test
+    void shouldSaveJudge() {
+        var adminData = createAdminData(judgeCode, judgeName, "ABERDEEN", "SALARIED");
+        var addJudgeService = new JudgeService(judgeRepository);
+        when(judgeRepository.existsByCodeOrName(judgeCode, judgeName)).thenReturn(false);
+        Assert.assertEquals(addJudgeService.saveJudge(adminData), true);
```
marcjudge, line 30: Swap the order of assertion here

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/admin/JudgeServiceTest.java
```diff
@@ -0,0 +1,60 @@
...
+        verify(judgeRepository, times(1)).save(createJudge(adminData));
+    }
+
+    @Test
+    void shouldReturnFalseIfJudgeExists() {
+        var adminData = createAdminData(judgeCode, judgeName, "ABERDEEN", "SALARIED");
+        var addJudgeService = new JudgeService(judgeRepository);
+        when(judgeRepository.existsByCodeOrName(judgeCode, judgeName)).thenReturn(true);
+        Assert.assertEquals(addJudgeService.saveJudge(adminData), false);
```
marcjudge, line 39: Swap the order of assertion here

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/domain/repository/JudgeRepository.java
```diff
@@ -10,4 +10,6 @@
 @Repository
 public interface JudgeRepository extends JpaRepository<Judge, Integer> {
     List<Judge> findByTribunalOffice(TribunalOffice tribunalOffice);
+
+    boolean existsByCodeOrName(String code, String name);
```
hsjhita, line 14: I think this needs to include Tribunal office. For example, if the same name is across different offices and they both had different codes, this wouldn't add in the judge because the name is the same despite it being in a different office

hsjhita: Bristol - 001 - Robin
Leeds - 002 - Robin 

If you tried something like this, I think the Leeds judge wouldn't be added

robin-ieong: Thanks Harpreet, as discussed yesterday I have updated the logic to check "if same code AND office exists already, if not then add the judge", so name check is removed. Also updated it to return a better error message in the response entity as you have suggested.

hsjhita: Name check is still required. 2 Judges in the same office cannot have the same name

hsjhita (approved):

cindychmcts (approved):

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/admin/staff/judge/JudgeServiceTest.java
```diff
@@ -0,0 +1,80 @@
...
+        var adminData = createAdminData(judgeCode, judgeName, tribunalOffice, "SALARIED");
+        var addJudgeService = new JudgeService(judgeRepository);
+        when(judgeRepository.existsByCodeAndTribunalOffice(judgeCode,TribunalOffice.valueOfOfficeName(tribunalOffice))).thenReturn(false);
+        when(judgeRepository.existsByNameAndTribunalOffice(judgeName,
+                TribunalOffice.valueOfOfficeName(tribunalOffice))).thenReturn(true);
+        assertThrows(SaveJudgeException.class, () -> addJudgeService.saveJudge(adminData));
+        verify(judgeRepository, never()).save(createJudge(adminData));
+    }
+
+    private AdminData createAdminData(String judgeCode, String judgeName, String tribunalOffice, String employmentStatus) {
```
petermoores-version1, line 63: Please refactor to introduce a Judge class passed around between methods, instead of many primitive type params.

petermoores-version1 (approved):

PR #322, opened 2022-06-29
RET-1985: Added ET3 Vetting Respondent Name and Address pages

JackReevies:

https://tools.hmcts.net/jira/browse/RET-1985
https://tools.hmcts.net/jira/browse/RET-1986
https://tools.hmcts.net/jira/browse/RET-1987
https://tools.hmcts.net/jira/browse/RET-1988

### Change description ###

Added callback to facilitate the Respondent Name and Address pages

```
[ ] Yes
[x] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/Et3VettingHelper.java
```diff
@@ -59,13 +67,73 @@ public static List<String> populateRespondentDynamicList(CaseData caseData) {
...
+            ? NO
+            : YES);
+
+        caseData.setEt3DoWeHaveRespondentsAddress(respondentAddress.equals(NONE_GIVEN)
+            ? NO
+            : YES);
+    }
+
+    public static String toAddressWithTab(Address address) {
+        StringBuilder claimantAddressStr = new StringBuilder();
```
hsjhita, line 124: refactor variable name

Brogie (approved):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/Et3VettingHelper.java
```diff
@@ -59,13 +67,73 @@ public static List<String> populateRespondentDynamicList(CaseData caseData) {
...
         List<DynamicValueType> dynamicRespondentList =
-                DynamicListHelper.createDynamicRespondentName(respondentCollection);
+            DynamicListHelper.createDynamicRespondentName(respondentCollection);
         DynamicFixedListType dynamicFixedListType = new DynamicFixedListType();
         dynamicFixedListType.setListItems(dynamicRespondentList);
         caseData.setEt3ChooseRespondent(dynamicFixedListType);
         return new ArrayList<>();
     }
 
+    public static void getRespondentNameAndAddress(CaseData caseData) {
```
nickwest1, line 77: Need JavaDoc on public methods.  A prime example here is that the method name doesn't describe what is actually happening and what happens if an address isn't found, etc...

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/Et3VettingHelper.java
```diff
@@ -156,29 +224,29 @@ private static boolean responseInTimePreValidationCheck(CaseData caseData, List<
...
+            ? "Cannot find ET1 Served Date"
+            : UtilHelper.listingFormatLocalDate(date);
     }
 
     private static String findEt3ExtensionDate(CaseData caseData) {
         List<RespondentSumTypeItem> respondentCollection = caseData.getRespondentCollection();
 
         if (CollectionUtils.isEmpty(respondentCollection)) {
-            log.error(NO_RESPONDENTS_FOUND_ERROR + caseData.getEthosCaseReference());
+            log.error(String.format(NO_RESPONDENTS_FOUND_ERROR, caseData.getEthosCaseReference()));
```
nickwest1, line 241: Good spot and good improvement ;)

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/RespondentBuilder.java
```diff
@@ -0,0 +1,71 @@
+package uk.gov.hmcts.ethos.replacement.docmosis.utils;
+
+import uk.gov.hmcts.et.common.model.ccd.Address;
+import uk.gov.hmcts.et.common.model.ccd.types.RespondentSumType;
+
+import static uk.gov.hmcts.ecm.common.model.helper.Constants.YES;
+
+public class RespondentBuilder {
```
nickwest1, line 8: Nice class, but we should have JavaDoc around it

ikirsanov: Cant we just create a class with fields and make a lomboks @Builder over it?

nickwest1: This is what I love about code reviews.  I never knew about the lambok @builder until you've mentioned it Ivan.  Nice suggestion :)
Jack, not sure if it's possible within the way you've built this, but certainly a great suggestion to make life easier in future.

JackReevies: Lombok builder looks really cool, correct me if I'm wrong but from what I've read it looks like the @Builder would go on the respondent object in data-model? We use @Data for this right now which does give us auto-generated get/setters (also really cool, saves a ton of time) but the advantage of using this builder here is that we can rollup multiple related properties into a single setter (for example `withExtension` takes in three related properties and sets them). I don't think @Builder is applicable in this scenario, but its definately something to keep in mind in future

ikirsanov: With builder you can build these 3 or 4 properties according to the situation and return this object and then add more fields and built it only in the final stage(absolute free to use with different situations), but maybe in this case it is more informative when you are marking what have you done within method.

ikirsanov (approved):

PR #385, opened 2022-07-21
RET-2191. Make checkstyle not ignore failures

Itay-Kirsh:
https://tools.hmcts.net/jira/browse/RET-2191

Make checkstyle failures not be ignored.

```
[ ] Yes
[x] No
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -22,6 +22,7 @@ allprojects {
 
 checkstyle {
     maxWarnings = 0
+    ignoreFailures = false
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -22,6 +22,7 @@ allprojects {
 
 checkstyle {
     maxWarnings = 0
+    ignoreFailures = false
```
gauravhmcts, line 25: Because we have severity set up as warning and not errors, will this setting be of any help?

PR #424, opened 2022-08-16
Ret 2205

hsjhita:

https://tools.hmcts.net/jira/browse/RET-2205

### Change description ###

Added doc gen for ET3 Vetting/Processing

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/DocumentHelper.java
```diff
@@ -961,4 +961,25 @@ public static Address getRespondentAddressET3(RespondentSumType respondentSumTyp
 
     }
 
+    /**
+     * Utility method to null check a string value.
+     * @param value to be checked if null or contains data
+     * @return null if the value is null or empty or will return the value itself
+     */
+    public static String nullChecker(String value) {
```
jthmcts, line 969: 1. Can't we reuse something instead? Is there no library that we can use to do this? What about StringUtils.defaultIfEmpty
2. I don't think DocumentHelper is an appropriate place for this method or listNullChecker. They are generic utility methods with nothing that ties them specifically to documents. They belong in a different class.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/Et3VettingService.java
```diff
@@ -90,4 +103,13 @@ private Et3VettingType getEt3VettingTypeForRespondent(RespondentSumTypeItem resp
     private Et3VettingType copyEt3FieldsFromCaseDataToRespondent(CaseData caseData) {
         return (Et3VettingType) intersectProperties(caseData, Et3VettingType.class);
     }
+
+    public DocumentInfo generateEt3ProcessingDocument(CaseData caseData, String userToken, String caseTypeId) {
```
jthmcts, line 107: Javadoc

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TornadoService.java
```diff
@@ -256,11 +258,22 @@ public DocumentInfo generateEt1VettingDocument(CaseData caseData, String userTok
...
+                case "ET1 Vetting.pdf" :
+                    outputStreamWriter.write(Et1VettingHelper.getDocumentRequest(caseData,
+                            tornadoConnection.getAccessKey()));
+                    break;
+                case "ET3 Processing.pdf" :
+                    outputStreamWriter.write(Et3VettingHelper.getDocumentRequest(caseData,
+                            tornadoConnection.getAccessKey()));
+                    break;
+                default:
+                    throw new NotFoundException("Document not found");
```
jthmcts, line 275: This is the wrong exception to throw. It's a webjars exception and nothing to do with what the code is doing. How about
`throw new IllegalArgumentException("Unexpected document name " + documentName);`

jthmcts: Also you should check the documentName argument at the start of the method

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/DocumentHelper.java
hsjhita: Replaced with defaultIsEmpty, generic utility methods removed from code

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/Et3VettingService.java
hsjhita: Javadoc added

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TornadoService.java
hsjhita: Corrected exception and added one at the start of the method - clarification on the duplication if possible?

hsjhita: I think I know what you mean. Both methods are creating the full request string but you think there's duplication over some properties? Out of the 4 top level fields, only 1 is the same - the access key, the rest change as it's specific to each document. I think that's what you mean, if not you'll need to clarify

hsjhita: Duplication removed

jthmcts (approved):

robin-ieong (approved):

PR #539, opened 2022-11-08
RET-2730: Add Unassigned to managingOffice

ikirsanov:
https://tools.hmcts.net/jira/browse/RET-2730

### Change description ###
For when we cannot find a managing office to assign to a case, which can occur when a postcode was not provided on the claimant work address or respondent address.

Managing office is added during case submission. If no office is added, as a valid postcode wasn't provided, we must assign it to the default office in the callbacks.

```
[ ] Yes
[ x] No
```


JackReevies (approved):

hsjhita (approved):
@ikirsanov I may have missed a PR but does the Unassigned office exist in the config files?

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TribunalOfficeServiceTest.java
```diff
@@ -47,8 +47,8 @@ public class TribunalOfficeServiceTest {
             { TribunalOffice.DUNDEE.getOfficeName(), "DD1 4QB" },
             { TribunalOffice.DUNDEE.getOfficeName(), "DD1 4QB" },
             { TribunalOffice.EDINBURGH.getOfficeName(), "EH3 7HF" },
-            { TribunalOffice.EDINBURGH.getOfficeName(), "EH3 7HF" },
-            { TribunalOffice.EDINBURGH.getOfficeName(), "EH3 7HF" }
+            { "Unassigned", "" },
```
GauravChawlaMOJ, line 50: Just a minor thing, may be use Unassigned constant here?

GauravChawlaMOJ (approved):

fergusoshea (approved):
Not an issue but for Line 24 doesn't "UNASSIGNED_OFFICE.equals(null)" already return false?

GauravChawlaMOJ:
> Not an issue but for Line 24 doesn't "UNASSIGNED_OFFICE.equals(null)" already return false?

@fergusoshea officeName will never be null in UNASSIGNED_OFFICE.equals(officeName) as it is being checked in the condition before. This condition is only checked if office is non-null. (OrElse/ ||)

ikirsanov:
> Not an issue but for Line 24 doesn't "UNASSIGNED_OFFICE.equals(null)" already return false?

If the value will be null, condition will not reach UNASSIGNED_OFFICE.equals(null) I think. If I will leave only UNASSIGNED_OFFICE.equals(value), then in case value == null, this If will be skipped and I dont need this. My situation is If the value is Null or The value is Unassigned then True.

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TribunalOfficeServiceTest.java
ikirsanov: Added constant to test

fergusoshea:
> If the value will be null, condition will not reach UNASSIGNED_OFFICE.equals(null) I think. If I will leave only UNASSIGNED_OFFICE.equals(value), then in case value == null, this If will be skipped and I dont need this. My situation is If the value is Null or The value is Unassigned then True.

You are right, I misread the second condition. My bad.

PR #717, opened 2023-02-16
Ret 3116

GauravChawlaMOJ:

https://tools.hmcts.net/jira/browse/RET-3116

### Change description ###

```
[ ] Yes
[X] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TornadoService.java
```diff
@@ -284,14 +285,15 @@ private void buildDocumentInstruction(HttpURLConnection connection, CaseData cas
...
+    private String getDocumentContent(CaseData caseData, String documentName, String caseTypeId, String userToken)
             throws JsonProcessingException {
         if (isNullOrEmpty(documentName)) {
             throw new NullPointerException("Document name cannot be null or empty");
         }
         switch (documentName) {
             case "ET1 Vetting.pdf":
-                return Et1VettingHelper.getDocumentRequest(caseData, tornadoConnection.getAccessKey());
+                return Et1VettingHelper.getDocumentRequest(caseData, tornadoConnection.getAccessKey(),
+                        reportDataService.getUserFullName(userToken));
```
hsjhita, line 296: correct me if I'm wrong but the way it's indicated in the configs is that the user types in their name but this will pull it down from IDAM. therefore you could have 2 different names?

hsjhita (approved):

PR #722, opened 2023-02-16
Ret 3206

GauravChawlaMOJ:
https://tools.hmcts.net/jira/browse/RET-3206

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
@@ -163,6 +163,9 @@ public class CreateReferralController {
         }
 
         CaseData caseData = ccdRequest.getCaseDetails().getCaseData();
+        if ("Party not responded/compiled".equals(caseData.getReferralSubject())) {
+            caseData.setReferralSubject("Party not responded/complied");
+        }
         UserDetails userDetails = userService.getUserDetails(userToken);
 
         emailService.sendEmail(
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
```diff
@@ -163,6 +163,9 @@ public ResponseEntity<CCDCallbackResponse> aboutToSubmitReferralDetails(
         }
 
         CaseData caseData = ccdRequest.getCaseDetails().getCaseData();
+        if ("Party not responded/compiled".equals(caseData.getReferralSubject())) {
+            caseData.setReferralSubject("Party not responded/complied");
```
hsjhita, line 167: I have a feeling this will require a new config entry

GauravChawlaMOJ: Can we hide that (duplicated) entry from the subjects list?

hsjhita: Can't hide fixed lists. I think as long as the label is showing the right spelling, it should be ok

GauravChawlaMOJ: As spoken to Harpreet, we would need this change for frontend to show correct value in the tab

btensay (approved):

PR #736, opened 2023-02-21
RET-3006

cindychmcts:

https://tools.hmcts.net/jira/browse/RET-3006

### Change description ###

Provide Something Tribunal has asked for- Common: Select a Judgement, Order, or Notification to view (MyHMCTS)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/PseRespondToTribunalController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/PseRespondToTribunalController.java
@@ -76,6 +76,40 @@ public class PseRespondToTribunalController {
     }
 
     /**
+     *  Populates the dynamic list for select an order or request to respond to.
+     *
+     * @param ccdRequest holds the request and case data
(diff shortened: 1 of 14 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/PseRespondToTribunalController.java
```diff
@@ -75,6 +75,40 @@ public ResponseEntity<CCDCallbackResponse> aboutToStart(
         return getCallbackRespEntityNoErrors(caseData);
     }
 
+    /**
+     *  Populates the dynamic list for select an order or request to respond to.
```
JackReevies, line 79: This javadoc is a bit confusing as its identical to the method above it. I see that one is populating select dropdown whereas the other one is select dropdown view but this doesn't really explain what/where the page is. If these are seperate events then it might be better to have a different controller file?

ikirsanov (approved):

fergusoshea (approved):

PR #752, opened 2023-02-28
New branch of ret-3172 implementation

btensay:
https://tools.hmcts.net/jira/browse/RET-3172

```
[ ] Yes
[ X] No
```


charts/et-cos/Chart.yaml
```diff
@@ -7,5 +7,5 @@ maintainers:
   - name: HMCTS Employment Tribunals Team
 dependencies:
   - name: java
-    version: 4.0.11
```
GauravChawlaMOJ, line 10: I think we need to update the version 0.0.21 to 0.0.22 too when we update the dependencies.

btensay: Thanks. I do that.

GauravChawlaMOJ (approved):

PR #861, opened 2023-04-06
Fix getSelectedListing

cindychmcts:
### Change description ###

Fix getSelectedListing

```
[ ] Yes
[x] No
```


charts/et-cos/Chart.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: et-cos
 home: https://github.com/hmcts/et-ccd-callbacks
-version: 0.0.23
+version: 0.0.24
```
GauravChawlaMOJ, line 4: Do we need to update chart version?

fergusoshea (approved):

PR #863, opened 2023-04-11
log messages added

btensay:
Referral and Referral reply emails log messages added.

https://tools.hmcts.net/jira/browse/RET-3367

```
[ ] Yes
[ X ] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
@@ -179,11 +179,10 @@ public class CreateReferralController {
                 userDetails.getName()
             )
         );
-
+        
         log.info("Event: Referral Email sent. "
             + "EventId: " + ccdRequest.getEventId() + "."
-            + "Emailed to: " + caseData.getReplyToEmailAddress() + " at : " + DateTime.now()
-            + " and sent by " + userDetails.getName() + ".");
+            + "Emailed at: " + DateTime.now());
 
         caseData.setReferredBy(String.format("%s %s", userDetails.getFirstName(), userDetails.getLastName()));
         DocumentInfo documentInfo = createReferralService.generateCRDocument(caseData,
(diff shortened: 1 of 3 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
```diff
@@ -179,6 +180,11 @@ public ResponseEntity<CCDCallbackResponse> aboutToSubmitReferralDetails(
             )
         );
 
+        log.info("Event: Referral Email sent. "
+            + "EventId: " + ccdRequest.getEventId() + "."
+            + "Emailed to: " + caseData.getReplyToEmailAddress() + " at : " + DateTime.now()
+            + " and sent by " + userDetails.getName() + ".");
```
hsjhita, line 186: we shouldn't mention the email address and user name as it's personal information. We shouldn't be have names in our logs

hsjhita (changes requested):
we shouldn't mention the email address and user name as it's personal information. We shouldn't be having names in our logs

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
btensay: Fair point.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/CreateReferralController.java
```diff
@@ -178,6 +179,10 @@ public ResponseEntity<CCDCallbackResponse> aboutToSubmitReferralDetails(
                 userDetails.getName()
             )
         );
+        
+        log.info("Event: Referral Email sent. "
```
GauravChawlaMOJ, line 183: not sure if it will help in diagnosing the issue of duplicate emails. May be worth checking if a loading/processing can be impmeneted here like RET-3435, in case user clicks the button twice

btensay: Thanks. I have a look.

btensay: I had a look and realised that this will change the scope of the ticket. I think from the ticket description, the ticket is intended to help understand what is going wrong and not directly to fix it.

btensay: Also, the submit button disabling feature that shows response to the user's first click on submit button for the first time, already exists. I tested it on perftest. It does gray out the button + display the "loading" animation.

GauravChawlaMOJ (approved):

PR #934, opened 2023-05-11
Ret 3253 Apps & Case Management common-View an application/case judgment from Case overview page

paul-peasegood:
https://tools.hmcts.net/jira/browse/RET-3253

### Change description ###

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseService.java
```diff
@@ -248,6 +251,45 @@ public String formatApplicationResponses(GenericTseApplicationType application,
...
+        ));
+
+        return MarkdownHelper.createTwoColumnTable(new String[]{"Decision", ""}, rows);
+    }
+
+    private String formatApplicationDecision(GenericTseApplicationType application, String authToken) {
+
+        String decisionsMarkdown = "";
+        if (application.getAdminDecision() != null) {
+            // Multiple decisions can be made for the same application but we are only showing the last one for now
```
JackReevies, line 280: Why are we only showing the last?  This doesn't look accurate given the code below it?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseService.java
```diff
@@ -248,6 +251,45 @@ public String formatApplicationResponses(GenericTseApplicationType application,
...
+            new String[]{"Sent to", decision.getSelectPartyNotify()}
+        ));
+
+        return MarkdownHelper.createTwoColumnTable(new String[]{"Decision", ""}, rows);
+    }
+
+    private String formatApplicationDecision(GenericTseApplicationType application, String authToken) {
+
+        String decisionsMarkdown = "";
+        if (application.getAdminDecision() != null) {
```
JackReevies, line 279: Can this if statement be inversed? Instead of doing something IF a condition is false, can we return early if the condition is true?

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseAdmCloseServiceTest.java
```diff
@@ -98,21 +98,7 @@ void generateCloseApplicationMarkdown(boolean appHasDoc, boolean appHasDetails,
...
-            + "|Date | 23 December 2022|\r\n"
-            + "|Sent by | Tribunal|\r\n"
-            + "|Type of decision | type of decision|\r\n"
-            + (hasAdditionalInfo ? "|Additional information | additional info|\r\n" : "")
-            + (hasDoc ? "|Document|<a href=\"/documents/%s\" target=\"_blank\">document (TXT, 1MB)</a>|\r\n" : "")
-            + "|Decision made by | decision made by|\r\n"
-            + "|Name | made by full name|\r\n"
-            + "|Sent to | party notify|\r\n"
-            + "\r\n";
+        String expected = "Application Details\r\n";
```
JackReevies, line 101: Is this test necessary now? It's being run 3 times with different values, but they don't seem to be needed

JackReevies (changes requested):

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseAdmCloseServiceTest.java
paul-peasegood: I guess not, should I remove it?

JackReevies: Sorry that was worded badly, a single test is valid but it doesn't need to be run 3 times now by the looks of it

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseService.java
paul-peasegood: Just checked with Boye, he wants just the recent decision. Ran a manual test it only shows most recent

JackReevies: Ah I see now, the `.reduce` part is responsible for getting the last item, is this a little convoluted? Would something like this be better?

```
var decisions = application.getAdminDecisions(); 
if (isEmpty(decisions)) {
  return "";
}
var lastItem = decisions.get(decisions,size() - 1);
var markdown = getSingleDecisionMarkdown(lastItem.getValue(), authToken);
```

JackReevies (commented):
removing request changes

brendanmchugh (approved):

ikirsanov (approved):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseService.java
```diff
@@ -248,6 +252,43 @@ public String formatApplicationResponses(GenericTseApplicationType application,
...
+            new String[]{"Additional information", decision.getAdditionalInformation()},
+            new String[]{"Decision made by", decision.getDecisionMadeBy()},
+            new String[]{"Name", decision.getDecisionMadeByFullName()},
+            new String[]{"Sent to", decision.getSelectPartyNotify()}
+        ));
+
+        return MarkdownHelper.createTwoColumnTable(new String[]{"Decision", ""}, rows);
+    }
+
+    private String formatApplicationDecision(GenericTseApplicationType application, String authToken) {
```
JackReevies, line 277:
```suggestion
    private String formatApplicationDecisions(GenericTseApplicationType application, String authToken) {
```

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseAdmCloseServiceTest.java
```diff
@@ -84,12 +78,11 @@ void updateStatusToClose() {
...
 
-    @ParameterizedTest
+    @Test
     @MethodSource("generateCloseApplicationMarkdown")
-    void generateCloseApplicationMarkdown(boolean appHasDoc, boolean appHasDetails,
-                                          boolean hasDoc, boolean hasAdditionalInfo) {
+    void generateCloseApplicationMarkdown() {
         GenericTseApplicationType tseApplicationType =
-            getTseAppType(appHasDoc, appHasDetails, hasDoc, hasAdditionalInfo);
+            getTseAppType(true, true, true, true);
```
JackReevies, line 85: This can be refactored. The params to this method are useless now, they don't do anything (intelliJ even says so). Can also remove the @'MethodSource tag

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseServiceTest.java
```diff
@@ -265,7 +267,71 @@ void formatViewApplication_withAllData() {
...
                 + "|Request made by|Caseworker|\r\n"
                 + "|Full name|Mr Lee Gal Officer|\r\n"
-                + "|Sent to|Respondent|\r\n\r\n";
+                + "|Sent to|Respondent|\r\n\r\n\r\n";
+
+        assertThat(tseService.formatViewApplication(caseData, AUTH_TOKEN)).isEqualTo(expected);
+    }
+
+    @Test
+    void formatViewApplication_withAllDataAndDecision() {
```
Itay-Kirsh, line 276: Fix the name of the previous method if it didn't have all data before. Also include multiple decisions to show it can handle it (have more than 2 with different dates and show they pick the latest 2)

Itay-Kirsh (changes requested):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/TseService.java
```diff
@@ -248,6 +252,43 @@ public String formatApplicationResponses(GenericTseApplicationType application,
...
+            return "";
+        }
+        List<String> decisionsMarkdown = application.getAdminDecision()
+            .stream()
+            .sorted(Comparator.comparing((TseAdminRecordDecisionTypeItem d) -> d.getValue().getDate()).reversed())
+            .limit(2)
+            .map(d -> getSingleDecisionMarkdown(d.getValue(), authToken))
+            .collect(Collectors.toList());
+
+        return String.join("\r\n", decisionsMarkdown);
```
JackReevies, line 289: Doesn't look like we have a test to make sure this code actually grabs the last two and concats them

JackReevies (changes requested):

PR #956, opened 2023-05-30
RET-3461 - cleaned up method & removed confusing Ternarys

aaronjduffy:
https://tools.hmcts.net/jira/browse/RET-3461

```
[ ] Yes
[x ] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/reports/timetofirsthearing/TimeToFirstHearingReport.java
```diff
@@ -294,40 +294,36 @@ private ReportSummary updateOpenTrack(ReportSummary reportSummary, boolean isFir
...
+        adhocReportType.setNotConFast26wkTotalPerCent(
+            calculatePercent(adhocReportType.getNotConFast26wkTotal(), conFastTotal));
+        adhocReportType.setNotConOpen26wkTotalPerCent(
+            calculatePercent(adhocReportType.getNotConOpen26wkTotal(), conOpenTotal));
+    }
+
+    private String calculatePercent(String value, int total) {
+        if (total == 0) {
+            return "0.00";
+        } else {
```
JackReevies, line 323: no need for an else clause here, the return line above will already quit out of this method if `total == 0`

JackReevies (changes requested):
great improvement here! Just one little minor comment

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/reports/timetofirsthearing/TimeToFirstHearingReport.java
```diff
@@ -294,40 +294,36 @@ private ReportSummary updateOpenTrack(ReportSummary reportSummary, boolean isFir
     }
 
     private void setPercent(AdhocReportType adhocReportType) {
```
Itay-Kirsh, line 296: This isn't tested at all so we really should add tests, verify that they pass on the previous code and on the new code as well.
As it's a private method (being called by other private methods higher up), go to the first public method that indirectly calls this and test it.

Itay-Kirsh (approved):

PR #1001, opened 2023-06-28
RET-3801 & RET-3800: Send emails to Claimant and Tribunal when LR res…

robin-ieong:

```
[ ] Yes
[x ] No
```


src/main/resources/application.properties
```diff
@@ -74,6 +74,8 @@ tse.respondent.respond.acknowledgement.rule92no.template.id=${TSE_RESPONDENT_ACK
 tse.respondent.respond.acknowledgement.rule92yes.template.id=${TSE_RESPONDENT_ACKNOWLEDGEMENT_YES_TEMPLATE_ID}
 tse.respondent.respond.notify.claimant.template.id=${TSE_RESPONDENT_RESPONSE_TEMPLATE_ID}
 tse.respondent.application.tribunal.template.id=${TSE_NEW_APPLICATION_ADMIN_TEMPLATE_ID}
+tse.respondent.reply-to-tribunal.to-tribunbal=${TSE_RESPONDENT_REPLY_TO_TRIB_TO_TRIB}
```
Itay-Kirsh, line 77: I suggest adding the values currently in -dev.properties directly to here as defaults. That means those with cftlib can run it automatically and we won't need the charts.values overrides or the overrides in the preview environment. This is what we're going to transition to soon enough anyway (as it is on et-sya-api).

robin-ieong: updated, thanks.

charts/et-cos/values.yaml
Itay-Kirsh, line 1: Can remove these now, they'll be picked up automatically from .properties

Itay-Kirsh (approved):

src/test/integration/resources/application.properties
```diff
@@ -68,3 +68,5 @@ url.exui.case-details=${EXUI_CASE_DETAILS_URL:dummyUrl}
 url.citizen.case-details=${CITIZEN_CASE_DETAILS_URL:dummyUrl}
 
 tse.respondent.application.tribunal.template.id=${TSE_NEW_APPLICATION_ADMIN_TEMPLATE_ID:42984f86-cbf3-4845-a0bd-2b0ea443e99d}
+tse.respondent.reply-to-tribunal.to-tribunbal=${TSE_RESPONDENT_REPLY_TO_TRIB_TO_TRIB:c2f72be8-541c-4bcc-b213-42557976dc39}
```
fergusoshea, line 71: Minor Typo "tribunbal" here and elsewhere this env var is used.

fergusoshea (approved):

src/test/integration/resources/application.properties
robin-ieong: thanks, fixed them

PR #1037, opened 2023-07-12
Add archive for functional and test reports

dharmendrak:


furnivall (changes requested):
Should we be adding the functional tests to the nightly jenkinsfile too? (i.e. not just CNP)

furnivall (approved):

hsjhita (approved):

PR #1042, opened 2023-07-17
RET-3883

GauravChawlaMOJ:

https://tools.hmcts.net/jira/browse/RET-3883

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseManagementForCaseWorkerService.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseManagementForCaseWorkerService.java
@@ -68,7 +68,7 @@ public class CaseManagementForCaseWorkerService {
     private static final String MESSAGE = "Failed to link ECC case for case id : ";
     private static final String CASE_NOT_FOUND_MESSAGE = "Case Reference Number not found.";
     public static final String LISTED_DATE_ON_WEEKEND_MESSAGE = "A hearing date you have entered "
-            + "falls on a weekend. You cannot list this case on a weekend. Please amend the date of Hearing ";
+        + "falls on a weekend. You cannot list this case on a weekend. Please amend the date of Hearing ";
 
     @Autowired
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseManagementForCaseWorkerService.java
```diff
@@ -195,6 +195,41 @@ public void setEt3ResponseDueDate(CaseData caseData) {
         }
     }
 
+    public void setNextListedDate(CaseData caseData) {
+        List<String> dates = new ArrayList<>();
+        String nextListedDate = "";
+
+        if (CollectionUtils.isNotEmpty(caseData.getHearingCollection())) {
+            for (HearingTypeItem hearingTypeItem : caseData.getHearingCollection()) {
+                dates = getListedDates(hearingTypeItem);
```
hsjhita, line 204: is this just not overriding dates every time with the latest result from getListedDates rather than adding them in all into a list?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseManagementForCaseWorkerService.java
```diff
@@ -195,6 +195,41 @@ public void setEt3ResponseDueDate(CaseData caseData) {
...
+
+            }
+        for (String date : dates) {
+            if (nextListedDate.equals("") && LocalDate.parse(date).isAfter(LocalDate.now())
```
hsjhita, line 208: would it make more sense to have a List<LocalDate> and parse the date when adding to the list rather than parsing it multiple times in this if statement?

hsjhita: and maybe a test to confirm the logic?

GauravChawlaMOJ: test added and date is parsed once only

hsjhita (approved):

PR #1102, opened 2023-08-17
additional changes added

btensay:
https://tools.hmcts.net/jira/browse/RET-3873

Additional changes added to match the updates made on template.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/TseHelperTest.java
+++ b/src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/TseHelperTest.java
@@ -227,7 +227,7 @@ class TseHelperTest {
             + "\",\"document_filename\":\"image.png\","
             + "\"document_url\":\"http://dm-store:8080/documents/1234\"},\"ownerDocument\":null,"
             + "\"creationDate\":null,\"shortDescription\":null}}],\"copy\":\"Yes\","
-            + "\"response\":\"response\",\"respondent\":\"Claimant\"}}";
+            + "\"response\":\"response\",\"respondentParty\":\"Claimant\"}}";
 
         assertThat(replyDocumentRequest, is(expected));
     }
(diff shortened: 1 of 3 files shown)


src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/TseHelperTest.java
```diff
@@ -220,14 +220,14 @@ void getReplyDocumentRequest_generatesData() throws JsonProcessingException {
...
-            + "\"type\":\"Withdraw my claim\",\"supportingYesNo\":\"Yes\","
+            + "\"type\":\"Withdraw my claim\",\"responseDate\":\"16-May-1996\",\"supportingYesNo\":\"Yes\","
             + "\"documentCollection\":[{\"id\":\"1234\","
             + "\"value\":{\"typeOfDocument\":null,"
             + "\"uploadedDocument\":{\"document_binary_url\":\"http://dm-store:8080/documents/1234/binary"
             + "\",\"document_filename\":\"image.png\","
             + "\"document_url\":\"http://dm-store:8080/documents/1234\"},\"ownerDocument\":null,"
             + "\"creationDate\":null,\"shortDescription\":null}}],\"copy\":\"Yes\","
-            + "\"response\":\"response\"}}";
+            + "\"response\":\"response\",\"respondent\":\"Claimant\"}}";
```
hsjhita, line 230: respondent = Claimant? is this not a bit confusing? Could this be called something else?

hsjhita: I understand why it's called respondent in some sense as it's the person respondent to the application but in our business terms, it might be confusing. If others think the same then I would suggest changing it

btensay: I kind of see what you mean. I renamed the field respondentParty and its values could be Claimant or Respondent as both make TSE response action.

hsjhita: Maybe something like responseFrom would've been better but I guess it works. You might also need to update the docmosis template as well

hsjhita (approved):

PR #1111, opened 2023-08-22
RET-3514 Removed the control for claimants who are not system users.

JustMehmet:

https://tools.hmcts.net/jira/browse/RET-3514

-----

### Change description ###

Removed the control for claimants who are not system users.

-----

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentTellSomethingElseController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentTellSomethingElseController.java
@@ -19,18 +19,15 @@ import uk.gov.hmcts.et.common.model.ccd.CCDRequest;
 import uk.gov.hmcts.et.common.model.ccd.CaseData;
 import uk.gov.hmcts.et.common.model.ccd.CaseDetails;
 import uk.gov.hmcts.et.common.model.ccd.items.GenericTseApplicationTypeItem;
-import uk.gov.hmcts.ethos.replacement.docmosis.helpers.Helper;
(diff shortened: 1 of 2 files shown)


hsjhita (approved):

btensay (approved):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentTellSomethingElseController.java
```diff
@@ -85,47 +82,10 @@ public ResponseEntity<CCDCallbackResponse> aboutToStartRespondentTSE(
...
-    @ApiResponses(value = {
-        @ApiResponse(responseCode = "200", description = "Accessed successfully",
-                content = {
-                    @Content(mediaType = "application/json",
-                            schema = @Schema(implementation = CCDCallbackResponse.class))
-                }),
-        @ApiResponse(responseCode = "400", description = "Bad Request"),
-        @ApiResponse(responseCode = "500", description = "Internal Server Error")
-    })
-    public ResponseEntity<CCDCallbackResponse> showError(
```
furnivall, line 113: does it make sense to keep this method (just remove the line about non system users?). Means it can be reused later if other errors need to be handled?

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentTellSomethingElseControllerTest.java
```diff
@@ -97,21 +91,6 @@ void afterEach() {
         mockHelper.close();
     }
 
-    @Test
```
furnivall, line 100: I don't think this test should be deleted (maybe changed?)

PR #1114, opened 2023-08-23
RET-4028-V2-Provide something else: Send a notification (ECM) - notify selected parties (email)

jip1984:
https://tools.hmcts.net/jira/browse/RET-4028

### Change description ###
When tribunal Sends out a notification for Other (General correspondence) / Hearing, the selected parties should receive an email notification to view the notification

```
[ ] Yes
[X ] No
```


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/SendNotificationService.java
```diff
@@ -176,16 +188,37 @@ public void sendNotifyEmails(CaseDetails caseDetails) {
...
-                buildPersonalisation(caseDetails, emailService.getCitizenCaseLink(caseId)));
+        if (!RESPONDENT_ONLY.equals(caseData.getSendNotificationNotify())) {
+
+            if (CollectionUtils.containsAny(caseData.getSendNotificationSubject(), SendNotificationSubjects)) {
+                emailService.sendEmail(sendNotificationTemplateId,
+                        claimantEmailAddress,
+                        buildPersonalisation(caseDetails, emailService.getCitizenCaseLink(caseId)));
+            }
+
+            if (CollectionUtils.containsAny(caseData.getSendNotificationSubject(),
```
marcjudge, line 199: Here and on L215, should these be else if? I'm not sure if we'd want to send more than one email even if the user has selected multiple.

src/main/resources/application.yaml
```diff
@@ -148,6 +148,10 @@ template:
   respondNotification:
     noResponse: ${RESPOND_NOTIFICATION_NO_RESPONSE_TEMPLATE_ID:d2caad87-2bab-4bc0-b598-01f1a7341f9e}
 
+  claimantSendNotificationHearingOther: ${CLAIMANT_SEND_NOTIFICATION_HEARING_OTHER_TEMPLATE_ID:a71ee3fd-a577-4a11-9bdb-de7fecd81336}
```
marcjudge, line 151: Please group these two with sendNotification template on L146

src/test/integration/resources/application.yaml
```diff
@@ -99,6 +99,10 @@ template:
   respondNotification:
     noResponse: d2caad87-2bab-4bc0-b598-01f1a7341f9e
 
+  claimantSendNotificationHearingOther: ${CLAIMANT_SEND_NOTIFICATION_HEARING_OTHER_TEMPLATE_ID:a71ee3fd-a577-4a11-9bdb-de7fecd81336}
```
marcjudge, line 102: Please group these two with sendNotification template on L97

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/SendNotificationService.java
jip1984: Having spoke with BA I think it is correct behaviour as hearing and other checkbox selected would send new template and if the rest of the checkbox values would send original template so in the case of for example hearing and claim(et1) were selected then both notification emails should be sent the original and the new one.

brendanmchugh (approved):

marcjudge (approved):

addison-jephtah (approved):

marcjudge (approved):

PR #1135, opened 2023-09-04
update logic to set respondent id

GauravChawlaMOJ:

https://tools.hmcts.net/jira/browse/RET-3884

### Change description ###

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocRespondentHelper.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocRespondentHelper.java
@@ -1,5 +1,6 @@
 package uk.gov.hmcts.ethos.replacement.docmosis.helpers;
 
+import org.elasticsearch.common.Strings;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.ecm.common.idam.models.UserDetails;
 import uk.gov.hmcts.et.common.model.bulk.types.DynamicFixedListType;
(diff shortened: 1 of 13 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocRespondentHelper.java
```diff
@@ -98,24 +98,15 @@ public RespondentSumType getRespondent(String respName, CaseData caseData) {
...
-        final List<RespondentSumTypeItem> respondentCollection = caseData.getRespondentCollection();
-        return respondentCollection.stream()
-                .filter(resp ->
-                        resp.getValue().getRespondentName()
-                                .equals(respondentRep.getValue().getRespRepName())).findFirst();
-    }
-
-    public List<RepresentedTypeRItem> updateWithRespondentIds(CaseData caseData) {
-        List<RepresentedTypeRItem> repList = new ArrayList<>();
+    public void updateWithRespondentIds(CaseData caseData) {
```
furnivall, line 101: Null checks are needed on this throughout. Also, can you avoid nested for loops somehow?

furnivall (changes requested):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocRespondentHelper.java
GauravChawlaMOJ: null check added, we can avoid nested loops by writing a stream.filter() which was already there, but it was changing the order of items in the collection in case the value doesn't exist. This way, it will not change the order.

furnivall (approved):

PR #1207, opened 2023-10-05
code change to add three new tabs

btensay:
(https://tools.hmcts.net/jira/browse/RET-4045)

### Change description ###
Three new tabs added for listing/displaying ADR Document, PII Document, and Appeal Document collections.

### Checklist

- [x] commit messages are meaningful and follow good commit message guidelines
- [x] tests have been updated / new tests has been added (if needed)

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/AdrDocumentsController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/AdrDocumentsController.java
@@ -1,0 +1,57 @@
+package uk.gov.hmcts.ethos.replacement.docmosis.controllers;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.media.Content;
+import io.swagger.v3.oas.annotations.media.Schema;
+import io.swagger.v3.oas.annotations.responses.ApiResponse;
+import io.swagger.v3.oas.annotations.responses.ApiResponses;
+import lombok.RequiredArgsConstructor;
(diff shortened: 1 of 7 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/AdrDocumentsController.java
```diff
@@ -0,0 +1,57 @@
...
+        @RequestBody CCDRequest ccdRequest,
+        @RequestHeader("Authorization") String userToken) {
+
+        if (!verifyTokenService.verifyTokenSignature(userToken)) {
+            log.error(INVALID_TOKEN, userToken);
+            return ResponseEntity.status(FORBIDDEN.value()).build();
+        }
+
+        return getCallbackRespEntityNoErrors(ccdRequest.getCaseDetails().getCaseData());
+    }
```
JustMehmet, line 56: This service turns only case data, looks not have any functionality. And all the other controller services does the same as turning back the case data?

Why did you need to implement the same code in 3 controllers?

btensay: The call removes empty tabs if the event fields are optional. Now they are mandatory, I can remove the controllers for now. Thanks

btensay: I had a logic in the controller method that puts upload date as fallback if it is not set in the UI. I had taken that out when I set the event fields mandatory.

JustMehmet (approved):

PR #1580, opened 2024-02-06
RET-4701: Added service for looking up cases

JackReevies:

https://tools.hmcts.net/jira/browse/RET-4701

### Change description ###

Added service for looking up cases

```
[] Yes
[x] No
```

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseLookupService.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseLookupService.java
@@ -21,18 +21,18 @@ public class CaseLookupService {
     private final AdminUserService adminUserService;
     private final CcdClient ccdClient;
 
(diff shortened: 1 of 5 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/CaseLookupService.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+@Slf4j
+@Service
+@RequiredArgsConstructor
+public class CaseLookupService {
+    public static final String EMPLOYMENT = "EMPLOYMENT";
```
hsjhita, line 20: Is it worth putting this in a Constants class? I'm pretty sure I've seen a lot of declarations for EMPLOYMENT and it makes sense for it to be a constant

hsjhita (approved):
LGTM

lvo700 (approved):

PR #1634, opened 2024-02-21
RET-4739

JustMehmet:
### Jira link (if applicable)

https://tools.hmcts.net/jira/browse/RET-4739

-----

### Change description ###

1. Updated to check all hearing type items if any of them has any hearing type which is disposed and has the same disposal date with the jurisdiction disposal date.

2. Added new empty environment variable, CFTLIB_EXTRA_COMPOSE_FILES to CFTLIB section of the build.gradle to not to fail CFTLIB config on local environments.

-----

### Checklist

- [ ] tests have been updated / new tests has been added (if needed)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
```diff
@@ -329,13 +330,54 @@ private void addInvalidDisposalDateError(List<HearingTypeItem> hearingTypeItems,
...
+     * Checks if hearing type items has any hearing type which is disposed
+     * and has the same disposal date with the jurisdiction disposal date.
+     * @param hearingTypeItems All hearings
+     * @param disposalDate Disposal date of Jurisdiction
+     * @return boolean  if true then hearing list has disposed hearing with the
+     *                  same disposal date of jurisdiction.
+     */
+    private boolean disposedHearingWithSameDisposalDateOfJurisdictionExists(
+            List<HearingTypeItem> hearingTypeItems, String disposalDate) {
+        if (ObjectUtils.isNotEmpty(hearingTypeItems)) {
```
hsjhita, line 348: is it worth inverting this condition to reduce complexity? not a big issue but just in terms of readability
if (CollectionUtils.isEmpty(hearingTypeItems) {
     return false;
 }

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
```diff
@@ -329,13 +330,54 @@ private void addInvalidDisposalDateError(List<HearingTypeItem> hearingTypeItems,
...
+
+    /**
+     * Checks if any of the dateListedTypeItems has any hearing type which is disposed
+     * and has the same disposal date with the jurisdiction disposal date.
+     * @param dateListedTypeItems All dateListedType Items in a hearing
+     * @param disposalDate Disposal date of Jurisdiction
+     * @return boolean  if true then at least one of the listed type itemshas
+     *                  disposed hearing with the same disposal date of jurisdiction.
+     */
+    private boolean hearingDateCollectionHasDisposedHearingWithSameDisposalDateOfJurisdiction(
```
hsjhita, line 369: method name is really long

hsjhita (approved):
Approved but have left some comments which can improve code quality. Also could maybe use streams instead of for loops

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
JustMehmet: I use long method names to make them understandable to everyone. It is a polemic between developers, If this is not the practice that we generally use then I will make it shorter. https://softwareengineering.stackexchange.com/questions/409455/clean-code-long-names-instead-of-comments

hsjhita (changes requested):
Also unit tests

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
JustMehmet: I am shortening the method names because in our projects we didn't use long method names. Thanks....

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
```diff
@@ -13,13 +13,7 @@
...
 import uk.gov.hmcts.et.common.model.ccd.items.RepresentedTypeRItem;
 import uk.gov.hmcts.et.common.model.ccd.items.RespondentSumTypeItem;
-import uk.gov.hmcts.et.common.model.ccd.types.CorrespondenceScotType;
-import uk.gov.hmcts.et.common.model.ccd.types.CorrespondenceType;
-import uk.gov.hmcts.et.common.model.ccd.types.HearingType;
-import uk.gov.hmcts.et.common.model.ccd.types.JudgementType;
-import uk.gov.hmcts.et.common.model.ccd.types.JurCodesType;
-import uk.gov.hmcts.et.common.model.ccd.types.RespondentSumType;
-import uk.gov.hmcts.et.common.model.ccd.types.RestrictedReportingType;
+import uk.gov.hmcts.et.common.model.ccd.types.*;
```
JackReevies, line 16: Checkstyle will not like this

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
```diff
@@ -329,13 +324,60 @@ private void addInvalidDisposalDateError(List<HearingTypeItem> hearingTypeItems,
...
+        if (CollectionUtils.isEmpty(hearingTypeItems)) {
+            return false;
+        }
+        for (HearingTypeItem hearingTypeItem : hearingTypeItems) {
+            HearingType hearingType = hearingTypeItem.getValue();
+            if (ObjectUtils.isEmpty(hearingType) || CollectionUtils.isEmpty(hearingType.getHearingDateCollection())) {
+                continue;
+            }
+            if (hasDisposedHearingWithSameDisposalDateOfJurisdiction(
+                    hearingType.getHearingDateCollection(), disposalDate)) {
```
JackReevies, line 351: Sorry to be pedantic but does this fit on a single line now?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
hsjhita: +1

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationService.java
JustMehmet: It fits Harpreet :). Already implemented it to fit on one column but I think it is showing like this on git :)

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EventValidationServiceTest.java
```diff
@@ -404,6 +409,23 @@ void shouldValidateJurisdictionCodesWithDuplicatesCodesAndExistenceJudgement() {
...
+    }
+
+    @Test
+    void shouldHaveValidationErrorForJurisdictionCodesOfDisposedHearingWithDifferentDisposalDateOfJurisdiction() {
+        List<String> errors = new ArrayList<>();
+        eventValidationService.validateJurisdiction(caseDetails24.getCaseData(), errors);
+
+        assertEquals(1, errors.size());
+        assertEquals(String.format(DISPOSAL_DATE_HEARING_DATE_MATCH, "DDA"), errors.get(0));
+    }
```
JackReevies, line 427: Thank you for adding the tests. I'd prefer to see you re-using existing case data and making the exact modifications within the test itself to make the method throw an error. Looking at these two tests alone its hard to see what input data changed that would lead to different output.

To see what exactly changed, I ran those two files through a diffchecker and there's only 1 field difference between them. Instead of adding a `caseDetailsTest24.json` file, It would be more appropriate here to re-use `caseDetailsTest23.json` and modify that field within the test function itself (like `caseData.getJurCodesCollection.get(0).getValue().setDisposalDate("2024-02-26")`).

But please compare this with one of the other test case details files, perhaps `caseDetailsTest23.json` is similar enough to `caseDetailsTest22.json` or another one?

JustMehmet: Thanks @JackReevies as we talked on the slack it is resolved by removing caseDetailsTest24.json. Unfortunately other existing json files doesn't match with the requirements that is why I was unable to use them. I agree with you that there should be only one json file and we should change the object values in test methods. But it would take a long time to correct it. That is why we need to handle it on a new task.

JackReevies (approved):

JackReevies (approved):

hsjhita (approved):

PR #1702, opened 2024-03-08
RET-0000: Added request interceptor for token verify service

JackReevies:

https://tools.hmcts.net/jira/browse/RET-0000

### Change description ###

Added request interceptor for token verify service - like what we have for et-sya-api

This doesn't remove the explicit checks from controllers - trying to keep PRs small

```
[] Yes
[x] No
```


hsjhita (approved):

hsjhita:
Really good work @JackReevies! Hopefully we can start removing it from the controllers in the main package!

jip1984 (approved):

PR #2716, opened 2025-07-29
Refactor vexationCheck method to use _source to help improve performance

hsjhita:
Refactor vexationCheck method to use _source to help improve performance by only searching for a subset of fields

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/Et1SubmissionService.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/Et1SubmissionService.java
@@ -27,6 +27,7 @@ import java.util.ArrayList;
 import java.util.Comparator;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.ForkJoinPool;
 import java.util.stream.IntStream;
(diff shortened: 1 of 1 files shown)


btensay (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/Et1SubmissionService.java
```diff
@@ -272,25 +267,50 @@ private static String findLanguagePreference(CaseData caseData) {
...
      * @param userToken the user token
      */
     public void vexationCheck(CaseDetails caseDetails, String userToken) {
-        String query = new SearchSourceBuilder()
-            .size(MAX_ES_SIZE)
-            .query(boolQuery()
-                .mustNot(new TermsQueryBuilder("state.keyword", "AWAITING_SUBMISSION_TO_HMCTS"))
-                .filter(new RangeQueryBuilder("data.receiptDate").gte(LocalDate.now().minusMonths(6L))))
-            .toString();
+        String query = """
```
LahiruOshara, line 276: can't we use SearchSourceBuilder and QueryBuilders to construct this ES query. will be more readable

hsjhita: @LahiruOshara the reason I'm not using SearchSourceBuilder is due to what I think is a CCD issue. When I tried expanding the original solution to include _source, the way the library handles it and outputs it is like below. CCD didn't like this which is why I've got the query in a raw string format

```
  "_source": {
    "includes": [
      "data.ethosCaseReference, data.claimant, state, data.receiptDate, data.respondent"
    ],
    "excludes": [ ]
  }
```

LahiruOshara (approved):

PR #2851, opened 2025-09-12
Shared list emails

LahiruOshara:
### Change description

- ET1 serving emails will be send to the legal reps in the shared list
- Applications acknowledgement emails will be send to the claimant legal reps

### Security Vulnerability Assessment ###

**CVE Suppression:** Are there any CVEs present in the codebase (either newly introduced or pre-existing) that are being intentionally suppressed or ignored by this commit?
  * [ ] Yes
  * [ ] No

### Checklist

- [ ] commit messages are meaningful and follow good commit message guidelines
- [ ] tests have been updated / new tests has been added (if needed)

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/applications/claimant/ClaimantTellSomethingElseController.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/applications/claimant/ClaimantTellSomethingElseController.java
@@ -122,7 +122,7 @@ public class ClaimantTellSomethingElseController {
         if (Helper.isRespondentSystemUser(caseData)) {
             claimantTseService.sendRespondentsEmail(caseDetails);
         }
-        claimantTseService.sendAcknowledgementEmail(caseDetails, userToken);
+        claimantTseService.sendAcknowledgementEmail(caseDetails);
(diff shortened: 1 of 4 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EmailNotificationService.java
```diff
@@ -0,0 +1,43 @@
...
+            if (ClaimantSolicitorRole.CLAIMANTSOLICITOR.getCaseRoleLabel().equals(assignment.getCaseRole())) {
+                UserDetails userDetails = adminUserService.getUserDetails(adminUserService.getAdminUserToken(),
+                        assignment.getUserId());
+                emailAddresses.add(userDetails.getEmail());
+            }
+        }
+        return emailAddresses;
+    }
+
+    public List<String> getRespondentSolicitorEmails(List<CaseUserAssignment> assignments) {
```
hsjhita, line 32: Rather than having a List, should probably change this to a Set just to handle duplicate values. if a solicitor is SOLICITORA and SOLICITORB, current solution will store the emails twice and will send out 2 of the same emails. Using a Set prevents duplicates so even if they are representing more than 1 respondent, they only get 1 email

PR #2916, opened 2025-11-25
RET-6074

JustMehmet:
### Jira link

https://tools.hmcts.net/jira/browse/RET-6074

-----

### Change description

Updates to resolve respondent representative removal bugs.

-----

### Test links

MyHMCTS/ExUI - https://xui-et-cos-pr-2916.preview.platform.hmcts.net/
CCD Admin Web - https://admin-web-et-cos-pr-2916.preview.platform.hmcts.net/
Camunda - https://camunda-et-cos-pr-2916.preview.platform.hmcts.net/
SYA Web - https://et-sya-et-cos-pr-2916.preview.platform.hmcts.net/
SYR Web - https://et-syr-et-cos-pr-2916.preview.platform.hmcts.net/
Manage Org - https://xui-mo-webapp-et-cos-pr-2916.preview.platform.hmcts.net/
Register Org - https://xui-mo-webapp-et-cos-pr-2916.preview.platform.hmcts.net/register-org/register
Approve Org - https://xui-ao-webapp-et-cos-pr-2916.preview.platform.hmcts.net/

### Checklist


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +61,134 @@ public class RespondentRepresentativeController {
...
+            @RequestHeader(AUTHORIZATION) String userToken) {
+        log.info("AMEND RESPONDENT REPRESENTATIVE SUBMITTED ---> " + LOG_MESSAGE + "{}",
+                callbackRequest.getCaseDetails().getCaseId());
+        try {
+            NocUtils.validateCallbackRequest(callbackRequest);
+            nocRespondentRepresentativeService.removeOldRepresentatives(callbackRequest, userToken);
+            nocRespondentRepresentativeService.addNewRepresentatives(callbackRequest);
+            nocRespondentRepresentativeService
+             .removeClaimantRepresentativeIfOrganisationExistsInRespondent(callbackRequest.getCaseDetails());
+            // nocRespondentRepresentativeService.updateRespondentRepresentativesAccess(callbackRequest);
```
hsjhita: Can this commented out code be removed?
```suggestion
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,724 @@
...
+                    nocWarnings.append(warningMessage).append('\n');
+                }
+            } catch (Exception e) {
+                // for localhost if e-mail is not entered same as wiremock request
+                String warningMessage = String.format(WARNING_REPRESENTATIVE_ACCOUNT_NOT_FOUND_BY_EMAIL,
+                        representativeName, representativeEmail);
+                nocWarnings.append(warningMessage).append('\n');
+            }
+        }
+        caseData.setNocWarning(nocWarnings.toString());
```
hsjhita: This field is only being set and the value is not being used anywhere in the code base. I assume this needs to be returned to be used in the controller?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocRespondentHelper.java
```diff
@@ -98,16 +106,99 @@ public RespondentSumType getRespondent(String respName, CaseData caseData) {
...
+
+        for (RespondentSumTypeItem respondent : caseData.getRespondentCollection()) {
+            if (!RespondentUtils.isValidRespondent(respondent)) {
+                continue;
+            }
+
+            String respondentId = respondent.getId();
+            String respondentName = respondent.getValue().getRespondentName();
+
+            boolean hasMatchingRepresentative = caseData.getRepCollection().stream()
```
hsjhita: Catering for null rep collection
```suggestion
            boolean hasMatchingRepresentative = emptyIfNull(caseData.getRepCollection()).stream()
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
```diff
@@ -0,0 +1,636 @@
...
+            String caseId = ObjectUtils.isEmpty(caseDetails) ? StringUtils.EMPTY : caseDetails.getCaseId();
+            log.warn(WARNING_INVALID_CASE_DETAILS_TO_NOTIFY_ORGANISATION_FOR_RESPONDENT_REP_UPDATE, caseId, nocType);
+            return;
+        }
+        if (!NotificationUtils.canNotifyRepresentativeOrganisation(representative)) {
+            log.warn(WARNING_INVALID_REPRESENTATIVE_TO_NOTIFY_ORGANISATION_FOR_RESPONDENT_REP_UPDATE,
+                    caseDetails.getCaseId(), nocType);
+            return;
+        }
+        String organisationId = representative.getValue().getRespondentOrganisation().getOrganisationID();
```
btensay: representative.getValue() needs null check

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/noc/RoleUtils.java
```diff
@@ -0,0 +1,397 @@
...
+                    SolicitorRole.SOLICITORF.getCaseRoleLabel()).organisation(Organisation.builder().build()).build());
+            case 6 -> caseData.setRespondentOrganisationPolicy6(OrganisationPolicy.builder().orgPolicyCaseAssignedRole(
+                    SolicitorRole.SOLICITORG.getCaseRoleLabel()).organisation(Organisation.builder().build()).build());
+            case 7 -> caseData.setRespondentOrganisationPolicy7(OrganisationPolicy.builder().orgPolicyCaseAssignedRole(
+                    SolicitorRole.SOLICITORH.getCaseRoleLabel()).organisation(Organisation.builder().build()).build());
+            case 8 -> caseData.setRespondentOrganisationPolicy8(OrganisationPolicy.builder().orgPolicyCaseAssignedRole(
+                    SolicitorRole.SOLICITORI.getCaseRoleLabel()).organisation(Organisation.builder().build()).build());
+            case 9 -> caseData.setRespondentOrganisationPolicy9(OrganisationPolicy.builder().orgPolicyCaseAssignedRole(
+                    SolicitorRole.SOLICITORJ.getCaseRoleLabel()).organisation(Organisation.builder().build()).build());
+            default -> { /* no-op */ }
```
hsjhita: Minor point but the readability of this isn't great. An idea could be in the OrganisationPolicy model, creating a method like

    public static OrganisationPolicy createDefaultPolicyByRole(String role) {
        return OrganisationPolicy.builder()
            .orgPolicyCaseAssignedRole(role)
            .organisation(Organisation.builder().build())
            .build();
    }

and the cases just becomes something like this

        case 0 -> caseData.setRespondentOrganisationPolicy0(
            createDefaultPolicyByRole(SolicitorRole.SOLICITORA.getCaseRoleLabel()));

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/ClaimantUtils.java
```diff
@@ -0,0 +1,44 @@
...
+                || StringUtils.isBlank(caseData.getClaimantType().getClaimantEmailAddress())) {
+            throw new NotFoundException(EXCEPTION_CLAIMANT_EMAIL_NOT_FOUND);
+        }
+        return caseData.getClaimantType().getClaimantEmailAddress();
+    }
+
+    public static String getClaimant(CaseData caseData) {
+        if (ObjectUtils.isEmpty(caseData)
+                || StringUtils.isBlank(caseData.getClaimant())) {
+            return StringUtils.EMPTY;
```
hsjhita: Rather than return empty, could we call the claimantDefaults method as that will set a value or a default if claimant is blank

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/LoggingUtils.java
```diff
@@ -0,0 +1,164 @@
...
+     * Only the exception message is logged, not the full stack trace.
+     *
+     * @param exception the exception originating from CCD whose message
+     *                  should be logged at INFO level
+     */
+    public static void logCcdErrorMessageAtInfoLevel(Exception exception) {
+        if (ObjectUtils.isEmpty(exception) || StringUtils.isBlank(exception.getMessage())) {
+            return;
+        }
+        log.info("Error form ccd - {}", exception.getMessage());
```
hsjhita: Typo
```suggestion
        log.info("Error from ccd - {}", exception.getMessage());
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
```diff
@@ -0,0 +1,695 @@
...
+            return StringUtils.EMPTY;
+        }
+        ResponseEntity<RetrieveOrgByIdResponse> organisationResponse = getOrganisationById(organisationId);
+        if (!NotificationUtils.canResolveOrganisationSuperuserEmail(caseDetails.getCaseId(), organisationId,
+                nocType, organisationResponse)) {
+            return StringUtils.EMPTY;
+        }
+        assert organisationResponse.getBody() != null;
+        return organisationResponse.getBody().getSuperUser().getEmail();
+    }
```
hsjhita: Method not being used
```suggestion
```

LahiruOshara (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocRespondentHelper.java
```diff
@@ -98,16 +106,77 @@ public RespondentSumType getRespondent(String respName, CaseData caseData) {
...
+     * exists that matches the respondent either by:
+     * <ul>
+     *     <li>respondent ID, or</li>
+     *     <li>respondent name</li>
+     * </ul>
+     * using the details present in the representative records.
+     * <p>
+     * A respondent is considered <em>unmatched</em> if no representative in the case has:
+     * <ul>
+     *     <li>a {@code respondentId} equal to the respondent's ID, nor</li>
```
btensay, line 123:
```suggestion
     *     <li>a {@code respondentId} equal to the respondent's ID, or</li>
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +69,268 @@ public class RespondentRepresentativeController {
...
+        log.info("AMEND RESPONDENT REPRESENTATIVE SUBMITTED ---> " + LOG_MESSAGE + "{}",
+                callbackRequest.getCaseDetails().getCaseId());
+        try {
+            NocUtils.validateCallbackRequest(callbackRequest);
+            nocRespondentRepresentativeService.updateRepresentativesAccess(callbackRequest, userToken);
+        } catch (GenericServiceException | GenericRuntimeException e) {
+            log.error(ERROR_UNABLE_TO_MODIFY_REPRESENTATIVE_ACCESS,
+                    callbackRequest.getCaseDetails().getCaseId(), e.getMessage());
+        }
+    }
```
btensay, line 201: **Just few thoughts**: 
Here we are quietly swallowing an error, i.e. returning 200 rather than 500 or so.  
Given the strict validation conditions and many subsequent conditional behaviours, if error/exception happens, I think, it is better to deal with it rather than give the user the impression of success. 
The visible error makes/forces the user to retry. In contrast, hiding the exception could cause inconsistency in case access.

config/owasp/suppressions.xml
```diff
@@ -81,4 +81,7 @@
     <suppress>
        <cve>CVE-2026-33929</cve>
     </suppress>
+    <suppress>
+       <cve>CVE-2026-42198</cve>
+    </suppress>
```
hsjhita, line 63: Suppression not needed
```suggestion
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/constants/NOCConstants.java
```diff
@@ -2,8 +2,153 @@
...
+    public static final String WARNING_INVALID_PARAMETERS_TO_RESOLVE_ORGANISATION_EMAIL =
+            "Invalid parameters(orgId, caseId, nocType). Unable to resolve organisation's superuser email Case id: {}";
+    public static final String WARNING_FAILED_TO_SEND_NOC_NOTIFICATION_EMAIL_ORGANISATION =
+            "Failed to send NOC notification email to organisation admin, case id: {}, error: {}";
+    public static final String WARNING_INVALID_CASE_DETAILS_TO_NOTIFY_TRIBUNAL_FOR_RESPONDENT_REP_UPDATE =
+            "Invalid case details. Unable to notify tribunal for respondent representative update. Case id: {}, "
+                    + NOC_TYPE;
+    public static final String WARNING_TRIBUNAL_EMAIL_NOT_FOUND_TO_NOTIFY_FOR_RESPONDENT_REP_UPDATE =
+            "Tribunal email not found. Unable to notify organisation for respondent representative update. "
+                    + "Case id: {},  NOC type: {}";
```
hsjhita: Just reusing the implementation you've put elsewhere
```suggestion
    public static final String WARNING_TRIBUNAL_EMAIL_NOT_FOUND_TO_NOTIFY_FOR_RESPONDENT_REP_UPDATE =
            "Tribunal email not found. Unable to notify organisation for respondent representative update. "
                    + "Case id: {},  " + NOC_TYPE;
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/MyHmctsService.java
```diff
@@ -59,7 +58,7 @@ public OrganisationAddress getOrganisationAddress(String userToken) throws Gener
                     ERROR_ORGANISATION_DETAILS_NOT_FOUND,
                     StringUtils.EMPTY,
                     "MyHmctsService",
-                    "getOrganisationAddress - organisation details not found");
+                    "getUseOrganisationAddress - organisation details not found");
```
hsjhita:
```suggestion
                    "getUserOrganisationAddress - organisation details not found");
```

LahiruOshara (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
JustMehmet: Thank you so much Tensay,
That controller service is called after submission of the event. In case of any error, we do not show it but log it. Even that exception is thrown we would not be able to show it and it will fall into retry mechanism. We had a conversation about this with Alex on the link https://hmcts-reform.slack.com/archives/D08MPV5DGGY/p1774285127814919 . 
Detailed message is as follows: 

I completely agree with your point. We should not be calling the method if the representative has already been removed.
To give a bit more context, the method
[removeOldRepresentatives](https://github.com/hmcts/et-ccd-callbacks/blob/cb8277a2e6a561865ee16c26281f527c128a9db8/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java#L185)
,
which is invoked by the respondent representative controller, first determines which representatives need to be removed. It does this via the utility method
[findRepresentativesToRemove](https://github.com/hmcts/et-ccd-callbacks/blob/cb8277a2e6a561865ee16c26281f527c128a9db8/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java#L229)
.
This method compares the old and new representatives (as lists) and checks whether their organisation or email address has changed. It also verifies whether a representative is eligible for removal (e.g. valid match, organisation change, or email change). Based on these criteria, it builds and returns a list of representatives to be removed.
The system then sends Notice of Change emails to those representatives and proceeds to revoke their access. During this step, it calls
[revokeOldRespondentRepresentativeAccess](https://github.com/hmcts/et-ccd-callbacks/blob/cb8277a2e6a561865ee16c26281f527c128a9db8/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java#L350)
,
which checks whether each removed representative has an existing case assignment. If a case assignment is found, their access is revoked and the OrganisationPolicy for the respondent representative is reset. If no assignment exists, no further action is taken.

However, the way we currently throw an exception isn’t an ideal solution for this scenario that is why I updated it to log an error. Because, even with additional checks in place, there’s still a risk of runtime exceptions, which would trigger the retry mechanism again.
Thank you so much Alex that was a great spot!.... (edited)

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocClaimantRepresentativeService.java
```diff
@@ -0,0 +1,288 @@
...
+     */
+    public String validateClaimantRepresentativeOrganisationMatch(CaseDetails caseDetails) {
+        String error = StringUtils.EMPTY;
+        if (ObjectUtils.isEmpty(caseDetails.getCaseData().getRepresentativeClaimantType())
+                || StringUtils.isBlank(caseDetails.getCaseData().getRepresentativeClaimantType()
+                .getRepresentativeEmailAddress())
+                || ObjectUtils.isEmpty(caseDetails.getCaseData().getRepresentativeClaimantType()
+                .getMyHmctsOrganisation())
+                || ObjectUtils.isEmpty(caseDetails.getCaseData().getRepresentativeClaimantType()
+                .getMyHmctsOrganisation().getOrganisationID())) {
```
hsjhita: Is it worth updating this method to use some of the util methods you created? The method before this (`validateRepresentativeOrganisationAndEmail`) has usage of those util classes
`hasHmctsOrganisationId`, `hasRepresentative` and `hasRepresentativeEmail`

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/noc/ClaimantRepresentativeUtils.java
```diff
@@ -0,0 +1,390 @@
...
+
+    private static void setRepresentativeId(RepresentedTypeC claimantRepresentative) {
+        if (claimantRepresentative == null) {
+            return;
+        }
+        if (StringUtils.isBlank(claimantRepresentative.getRepresentativeId())) {
+            claimantRepresentative.setRepresentativeId(UUID.randomUUID().toString());
+        }
+        if (ObjectUtils.isNotEmpty(claimantRepresentative.getMyHmctsOrganisation())
+                && StringUtils.isNotBlank(claimantRepresentative.getMyHmctsOrganisation().getOrganisationID())) {
```
hsjhita: Can this be changed to use the `hasHmctsOrganisationId` util method created

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/AmendRepresentativeContactController.java
```diff
@@ -0,0 +1,149 @@
...
+    })
+    public ResponseEntity<CCDCallbackResponse> aboutToSubmit(
+            @RequestBody CCDRequest ccdRequest,
+            @RequestHeader(HTTP_HEADER_AUTHORIZATION) String userToken) {
+        CaseData caseData = ccdRequest.getCaseDetails().getCaseData();
+        List<String> errors = new ArrayList<>();
+        try {
+            amendRepresentativeContactService.updateRepresentativeContactDetails(
+                    userToken, caseData, ccdRequest.getCaseDetails().getCaseId());
+            caseData.setMyHmctsAddressText(null);
```
hsjhita, line 143: Just nullifying the other field set in this event
```suggestion
            caseData.setMyHmctsAddressText(null);
            caseData.setEt3ResponseAddress(null);
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRepresentativeService.java
```diff
@@ -0,0 +1,220 @@
...
+    public CaseData updateRepresentation(CaseDetails caseDetails, String userToken) throws IOException {
+        CaseData caseData = caseDetails.getCaseData();
+        ChangeOrganisationRequest change = validateChangeRequest(caseData);
+        DynamicFixedListType caseRoleId = change.getCaseRoleId();
+
+        if (caseRoleId.getValue().getCode().equals(ClaimantSolicitorRole.CLAIMANTSOLICITOR.getCaseRoleLabel())) {
+            caseData = nocClaimantRepresentativeService.updateClaimantRepresentation(caseDetails, userToken);
+            caseDetails.setCaseData(caseData);
+            nocRespondentRepresentativeService.revokeRespondentRepresentativesWithSameOrganisationAsClaimant(
+                    caseDetails);
```
hsjhita, line 85: The email that is sent to respondents is the wrong email. The email template used suggests a rep is now representing a respondent but instead we are removing the rep from the respondent

Image

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1290 @@
...
+     *     <li>Respondent representatives belonging to the same organisation should not remain active.</li>
+     *     <li>Organisation policies must be reset after representatives are revoked.</li>
+     * </ul>
+     *
+     * @param caseDetails the case details containing case data and representative information
+     */
+    public void revokeRespondentRepresentativesWithSameOrganisationAsClaimant(CaseDetails caseDetails) {
+        if (ObjectUtils.isEmpty(caseDetails)
+                || StringUtils.isEmpty(caseDetails.getCaseId())
+                || ObjectUtils.isEmpty(caseDetails.getCaseData())) {
```
hsjhita: could this be replaced by the `validateCaseDetails` method in the `CaseDataUtils` class

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1290 @@
...
+     *
+     * @param caseDetails the case details containing case ID, type, jurisdiction, and existing case data
+     * @throws GenericRuntimeException if an error occurs while communicating with CCD services
+     */
+    public void resetOrganisationPolicies(CaseDetails caseDetails) {
+        if (ObjectUtils.isEmpty(caseDetails)
+                || ObjectUtils.isEmpty(caseDetails.getCaseData())
+                || StringUtils.isBlank(caseDetails.getCaseTypeId())
+                || StringUtils.isBlank(caseDetails.getJurisdiction())
+                || StringUtils.isEmpty(caseDetails.getCaseId())
```
hsjhita: Could some of this be replaced by the `validateCaseDetails` method in the `CaseDataUtils` class

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1290 @@
...
+            throw new GenericRuntimeException(exception);
+        }
+    }
+
+    private static @NonNull CaseData getCCDRequestCaseData(CaseDetails caseDetails, CCDRequest ccdRequest)
+            throws GenericServiceException {
+        final String methodName = "getCCDRequestCaseData";
+        if (ObjectUtils.isEmpty(ccdRequest)
+                || ObjectUtils.isEmpty(ccdRequest.getCaseDetails())
+                || StringUtils.isBlank(ccdRequest.getCaseDetails().getCaseId())) {
```
hsjhita, line 699: Could some of this be replaced by the `validateCCDRequest` method in the `CaseDataUtils` class

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1290 @@
...
+     * @param caseDetails the case details containing the case ID and case data to update
+     * @param representatives the respondent representatives to be granted access
+     */
+    public void grantRespondentRepresentativesAccess(CaseDetails caseDetails,
+                                                     List<RepresentedTypeRItem> representatives) {
+        if (ObjectUtils.isEmpty(caseDetails)
+                || ObjectUtils.isEmpty(caseDetails.getCaseData())
+                || StringUtils.isBlank(caseDetails.getCaseId())
+                || CollectionUtils.isEmpty(representatives)) {
+            return;
```
hsjhita: Could some of this be replaced by the `validateCaseDetails` method in the `CaseDataUtils` class

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1290 @@
...
+     * @param userToken   the IDAM user authentication token
+     * @param caseDetails the case details containing case ID and case data
+     * @return a list of {@link RepresentedTypeRItem} representing the respondent representatives
+     *         associated with the user; returns an empty list if none are found or inputs are invalid
+     */
+    public List<RepresentedTypeRItem> findRepresentativesByToken(String userToken, CaseDetails caseDetails) {
+        if (StringUtils.isBlank(userToken)
+                || ObjectUtils.isEmpty(caseDetails)
+                || StringUtils.isBlank(caseDetails.getCaseId())
+                || ObjectUtils.isEmpty(caseDetails.getCaseData())
```
hsjhita, line 963: Could some of this be replaced by the `validateCaseDetails` method in the `CaseDataUtils` class

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
hsjhita, line 1: With the Utils, I get that we might not want to throw the error but maybe the check in the utils class can be a boolean that can be used in this class as well as the utils class

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1290 @@
...
+     *         the existing {@link CaseData} if no action was taken or removal failed,
+     *         or {@code null} if {@code caseDetails} is {@code null}
+     */
+    public CaseData removeConflictingClaimantRepresentation(CaseDetails caseDetails) {
+        if (ObjectUtils.isEmpty(caseDetails)
+                || StringUtils.isEmpty(caseDetails.getCaseId())
+                || StringUtils.isBlank(caseDetails.getCaseTypeId())
+                || StringUtils.isBlank(caseDetails.getJurisdiction())
+                || ObjectUtils.isEmpty(caseDetails.getCaseData())
+                || CollectionUtils.isEmpty(caseDetails.getCaseData().getRepCollection())
```
hsjhita, line 918: Also caseDetails is checked in the If statement and the return

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
```diff
@@ -0,0 +1,693 @@
...
+        claimantEmailAddress = StringUtils.isBlank(claimantEmailAddress) ? StringUtils.EMPTY : claimantEmailAddress;
+        Map<String, String> personalisation = buildNoCPersonalisation(caseDetails, claimant);
+        personalisation.put(LINK_TO_CITIZEN_HUB, emailService.getCitizenCaseLink(caseDetails.getCaseId()));
+        try {
+            emailService.sendEmail(claimantTemplateId, claimantEmailAddress, personalisation);
+        } catch (Exception e) {
+            log.warn(WARNING_FAILED_TO_SEND_REMOVAL_OF_REPRESENTATIVE_CLAIMANT, caseDetails.getCaseId(),
+                    e.getMessage());
+        }
+    }
```
hsjhita: We should probably see if the claimant has access to the portal before sending them the email otherwise we could be sending someone an email for something they don't have access to.

Also if `claimantEmailAddress` if blank and is set to an empty string, we should probably have an early return rather than trying to send an email to a blank string

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +69,268 @@ public class RespondentRepresentativeController {
...
+        @ApiResponse(responseCode = HTTP_CODE_FOUR_ZERO_FOUR, description = HTTP_MESSAGE_FOUR_ZERO_FOUR),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_HUNDRED, description = HTTP_MESSAGE_FIVE_HUNDRED),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_ONE, description = HTTP_MESSAGE_FIVE_ZERO_ONE),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_THREE, description = HTTP_MESSAGE_FIVE_ZERO_THREE)
+    })
+    public ResponseEntity<CCDCallbackResponse> amendRespondentRepresentativeAboutToStart(
+            @RequestBody CCDRequest ccdRequest,
+            @RequestHeader(AUTHORIZATION) String userToken) {
+        CaseDataUtils.validateCCDRequest(ccdRequest);
+        log.info("AMEND RESPONDENT REPRESENTATIVE ABOUT TO START ---> " + LOG_MESSAGE + "{}",
```
hsjhita: Should the `LOG_MESSAGE` be part of this log message? Would log `AMEND RESPONDENT REPRESENTATIVE ABOUT TO START --->  received respondent's remove own representative request for case reference : 1111222233334444` for example


```suggestion
        log.info("AMEND RESPONDENT REPRESENTATIVE ABOUT TO START ---> {}",
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +69,268 @@ public class RespondentRepresentativeController {
...
+        @ApiResponse(responseCode = HTTP_CODE_FOUR_ZERO_FOUR, description = HTTP_MESSAGE_FOUR_ZERO_FOUR),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_HUNDRED, description = HTTP_MESSAGE_FIVE_HUNDRED),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_ONE, description = HTTP_MESSAGE_FIVE_ZERO_ONE),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_THREE, description = HTTP_MESSAGE_FIVE_ZERO_THREE)
+    })
+    public ResponseEntity<CCDCallbackResponse> amendRespondentRepresentativeMidEvent(
+            @RequestBody @NotNull CCDRequest ccdRequest,
+            @RequestHeader(AUTHORIZATION) String userToken) {
+        CaseDataUtils.validateCCDRequest(ccdRequest);
+        log.info("CHECKING RESPONDENT REPRESENTATIVE ORGANISATION ---> " + LOG_MESSAGE + "{}",
```
hsjhita: Same thing here
```suggestion
        log.info("CHECKING RESPONDENT REPRESENTATIVE ORGANISATION ---> {}",
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +69,268 @@ public class RespondentRepresentativeController {
...
+        @ApiResponse(responseCode = HTTP_CODE_FOUR_ZERO_FOUR, description = HTTP_MESSAGE_FOUR_ZERO_FOUR),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_HUNDRED, description = HTTP_MESSAGE_FIVE_HUNDRED),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_ONE, description = HTTP_MESSAGE_FIVE_ZERO_ONE),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_THREE, description = HTTP_MESSAGE_FIVE_ZERO_THREE)
+    })
+    public ResponseEntity<CCDCallbackResponse> amendRespondentRepresentativeAboutToSubmit(
+            @RequestBody @NotNull CCDRequest ccdRequest,
+            @RequestHeader(AUTHORIZATION) String userToken) {
+        CaseDataUtils.validateCCDRequest(ccdRequest);
+        log.info("AMEND RESPONDENT REPRESENTATIVE ABOUT TO SUBMIT ---> " + LOG_MESSAGE + "{}",
```
hsjhita:
```suggestion
        log.info("AMEND RESPONDENT REPRESENTATIVE ABOUT TO SUBMIT ---> {}",
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +69,268 @@ public class RespondentRepresentativeController {
...
+        @ApiResponse(responseCode = HTTP_CODE_FOUR_ZERO_THREE, description = HTTP_MESSAGE_FOUR_ZERO_THREE),
+        @ApiResponse(responseCode = HTTP_CODE_FOUR_ZERO_FOUR, description = HTTP_MESSAGE_FOUR_ZERO_FOUR),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_HUNDRED, description = HTTP_MESSAGE_FIVE_HUNDRED),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_ONE, description = HTTP_MESSAGE_FIVE_ZERO_ONE),
+        @ApiResponse(responseCode = HTTP_CODE_FIVE_ZERO_THREE, description = HTTP_MESSAGE_FIVE_ZERO_THREE)
+    })
+    public void amendRespondentRepresentativeSubmitted(
+            @RequestBody CallbackRequest callbackRequest,
+            @RequestHeader(AUTHORIZATION) String userToken) {
+        log.info("AMEND RESPONDENT REPRESENTATIVE SUBMITTED ---> " + LOG_MESSAGE + "{}",
```
hsjhita:
```suggestion
        log.info("AMEND RESPONDENT REPRESENTATIVE SUBMITTED ---> {}",
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/controllers/RespondentRepresentativeController.java
```diff
@@ -48,6 +69,268 @@ public class RespondentRepresentativeController {
...
+            @RequestHeader(AUTHORIZATION) String userToken) {
+        CaseDetails caseDetails = ccdRequest.getCaseDetails();
+        List<String> errors = new ArrayList<>();
+        try {
+            nocRespondentRepresentativeService.saveRespondentRepresentativeContactDetails(userToken, caseDetails);
+        } catch (GenericServiceException gse) {
+            errors.add(gse.getMessage());
+        }
+        return getCallbackRespEntityErrors(errors, caseDetails.getCaseData());
+    }
```
hsjhita: Are these needed now? Seems like `AmendRepresentativeContactController` replaces these?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocNotificationHelper.java
hsjhita: Given that this method is being used at line 121, would it better if it was named something more appropriate like `addCommonEmailValues`?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/helpers/NocNotificationHelper.java
```diff
@@ -113,11 +118,9 @@ public static Map<String, String> buildNoCPersonalisation(CaseDetails detail, St
     }
 
     public static Map<String, String> buildTribunalPersonalisation(CaseData caseData) {
-        Map<String, String> personalisation = new ConcurrentHashMap<>();
-
-        addCommonValues(caseData, personalisation);
-        personalisation.put(DATE, ReferralHelper.getNearestHearingToReferral(caseData, "Not set"));
-        personalisation.put("tribunal", isNullOrEmpty(caseData.getTribunalAndOfficeLocation()) ? UNKNOWN :
+        Map<String, String> personalisation = buildPreviousRespondentSolicitorPersonalisation(caseData);
```
hsjhita: See https://github.com/hmcts/et-ccd-callbacks/pull/2916/changes#r3350549756

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/noc/ClaimantRepresentativeUtilsTest.java
```diff
@@ -0,0 +1,354 @@
...
+    private static final String CLAIMANT_EMAIL_ADDRESS = "claimant@email.com";
+    private static final String CLAIMANT_REPRESENTATIVE_EMAIL_ADDRESS = "claimantrep@email.com";
+    private static final String ORGANISATION_ID_1 = "dummy12_organisation34_id56";
+    private static final String ORGANISATION_ID_2 = "dummy65_organisation43_id21";
+    private static final String RESPONDENT_REPRESENTATIVE_ID = "respondentRepresentativeId";
+    private static final String CLAIMANT_REPRESENTATIVE_ID = "claimantRepresentativeId";
+    private static final String CLAIMANT_REPRESENTATIVE_NAME = "claimantRepresentativeName";
+    private static final String ROLE_SOLICITORA = "[SOLICITORA]";
+    private static final String ROLE_SOLICITORB = "[SOLICITORB]";
+    private static final String ROLE_CLAIMANT_SOLICITOR = "[CLAIMANTSOLICITOR]";
```
hsjhita, line 42: I've noticed in some classes, you created these as public variables, some classes there are private; could we have a common approach or use the labels from the Enums?

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/CallbacksCollectionUtilsTest.java
```diff
@@ -0,0 +1,87 @@
...
+    public static final String ROLE_SOLICITOR_A = "[SOLICITORA]";
+    public static final String ROLE_SOLICITOR_B = "[SOLICITORB]";
+    public static final String ROLE_SOLICITOR_C = "[SOLICITORC]";
+    public static final String ROLE_SOLICITOR_D = "[SOLICITORD]";
+    public static final String ROLE_SOLICITOR_E = "[SOLICITORE]";
+    public static final String ROLE_SOLICITOR_F = "[SOLICITORF]";
+    public static final String ROLE_SOLICITOR_G = "[SOLICITORG]";
+    public static final String ROLE_SOLICITOR_H = "[SOLICITORH]";
+    public static final String ROLE_SOLICITOR_I = "[SOLICITORI]";
+    public static final String ROLE_SOLICITOR_J = "[SOLICITORJ]";
```
hsjhita: Relates to https://github.com/hmcts/et-ccd-callbacks/pull/2916/changes#r3354181299

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
hsjhita: @JustMehmet this method is still not used. If Cindy is using it in her PR, it should be an addition in her PR and not yours

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/NotificationUtils.java
```diff
@@ -0,0 +1,168 @@
...
+            String tmpCaseId = StringUtils.isBlank(caseId) ? StringUtils.EMPTY : caseId;
+            log.warn(WARNING_INVALID_PARAMETERS_TO_RESOLVE_ORGANISATION_EMAIL, tmpCaseId);
+            return false;
+        }
+        final String orgType = NOC_TYPE_REMOVAL.equals(nocType) ? OLD_LOWERCASE : NEW_LOWERCASE;
+        if (ObjectUtils.isEmpty(orgResponse)
+                || !orgResponse.getStatusCode().is2xxSuccessful()
+                || ObjectUtils.isEmpty(orgResponse.getBody())
+                || ObjectUtils.isEmpty(orgResponse.getBody().getSuperUser())
+                || StringUtils.isBlank(orgResponse.getBody().getSuperUser().getEmail())) {
```
hsjhita: reusing the method below
```suggestion
        if (hasOrganisationSuperuserEmail(orgResponse)) {
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/OrganisationService.java
```diff
@@ -0,0 +1,91 @@
...
+                String warningMessage = String.format(WARNING_REPRESENTATIVE_ACCOUNT_NOT_FOUND_BY_EMAIL,
+                        representativeName);
+                log.warn(warningMessage);
+                nocWarnings.add(warningMessage);
+            }
+        } catch (FeignException e) {
+            if (e.status() == HttpStatus.NOT_FOUND.value()) {
+                String warningMessage = String.format(WARNING_REPRESENTATIVE_ACCOUNT_NOT_FOUND_BY_EMAIL,
+                        representativeName);
+                log.warn(warningMessage);
```
hsjhita: In both these log messages, we could log out the representative's name which is exposing PII. Either the log message needs to be changed or removed

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
JustMehmet: Thanks for the feedback. I agree that generic utility methods should usually avoid hiding business-specific exception behaviour.

In this case, however, the method is intentionally a validation/guard method rather than a boolean utility check. The method name 'validateCaseDetails' is intended to make the fail-fast behaviour explicit: if the required case details are missing or invalid, it throws the service exception expected by the calling flow.

We already have/use boolean-style validation separately where needed, but here the callers need validation with the existing exception handling pattern. For that reason, I’d prefer to keep the current approach.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
JustMehmet: Thanks for the feedback. I’ve added an early return so no email is sent when the claimant email address is blank.

I haven’t added a portal access check based on the claimant email address, as the claimant email address on the case may have been updated and may not necessarily match the email address used for portal login/access. Checking access by this email could therefore incorrectly prevent the email from being sent to a valid claimant contact address.

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/noc/ClaimantRepresentativeUtilsTest.java
JustMehmet: Thanks for spotting this. These are only used in test classes, and the intention was to avoid reusing values/constants that are defined for actual production behaviour.

I agree that the approach should be consistent. I’ll move these test-only values into a shared `TestConstants` class in the next versions, and use those across the relevant test classes, rather than having a mix of public and private variables.

src/test/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/CallbacksCollectionUtilsTest.java
JustMehmet: Thanks for spotting this. These are only used in test classes, and the intention was to avoid reusing values/constants that are defined for actual production behaviour.

I agree that the approach should be consistent. I’ll move these test-only values into a shared TestConstants class in the next versions, and use those across the relevant test classes, rather than having a mix of public and private variables.
Updated all those variables as private.

charts/et-cos/values.preview.template.yaml
```diff
@@ -663,6 +663,7 @@ xui-webapp:
         secrets:
           - mc-s2s-client-secret
           - mc-idam-client-secret
+          - mc-session-secret
```
btensay: already exists on line 668

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
```diff
@@ -0,0 +1,694 @@
...
+     * </ul>
+     *
+     * <p>If any exception occurs while retrieving the claimant email address or sending
+     * the email, the error is logged and the method exits without throwing the exception
+     * further, ensuring the calling flow is not interrupted.
+     *
+     * @param caseDetails the {@link CaseDetails} containing the case and claimant information
+     *                    required to send the notification
+     */
+    public void notifyClaimantOfRepresentationRemoval(CaseDetails caseDetails) {
```
btensay: This method is called only from unit test and not exercised from code. Seems has become dead code due to the refactoring in this pr.  Please consider removing it and unit tests it has unless there is a compelling reason to keep it.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1273 @@
...
+                .getRepresentativeClaimantType());
+        if (StringUtils.isEmpty(organisationId)) {
+            return;
+        }
+        revokeAndRemoveRepresentativesByOrganisation(caseDetails, organisationId);
+    }
+
+    /**
+     * Finds respondent representatives associated with the given organisation,
+     * notifies the corresponding respondents, and delegates access revocation
```
btensay: Does the “notifies the corresponding respondents” bit still applies after the refactoring? This Javadoc line may need updating as well.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1271 @@
...
+                        representativeItem.getValue().getNameOfRepresentative(),
+                        representativeItem.getValue().getRepresentativeEmailAddress()));
+            }
+        }
+        return warnings;
+    }
+
+    public void updateRepresentativesAccess(CallbackRequest callbackRequest, String userToken) {
+        revokeOldRepresentatives(callbackRequest, userToken);
+        addNewRepresentatives(callbackRequest);
```
btensay, line 156: If revoking a rep fails, add new ones still goes ahead. From the point of view of giving the new rep access, it is ok. But the old reps may still remain with case access.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1271 @@
...
+                                remainingCaseUserAssignments, representative, caseUserAssignment.getCaseRole());
+                if (CollectionUtils.isNotEmpty(otherAssignmentsToRemove)) {
+                    caseUserAssignmentsToRevoke.addAll(otherAssignmentsToRemove);
+                }
+            }
+        }
+        try {
+            revokeCaseAssignments(userToken, caseUserAssignmentsToRevoke);
+        } catch (GenericRuntimeException e) {
+            log.error(ERROR_UNABLE_TO_NOTIFY_REPRESENTATION_REMOVAL, oldCaseDetails.getCaseId(), e.getMessage(), e);
```
btensay: Is this the correct error log? Should it not be about failed revoke case assignment? I think, notification related error is handled somewhere else.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -0,0 +1,1271 @@
...
+     * no action is taken and the existing {@link CaseData} is returned (or {@code null}
+     * if {@code caseDetails} itself is {@code null}).</p>
+     *
+     * <p>If validation passes, the method:
+     * <ol>
+     *     <li>Obtains an admin user token.</li>
+     *     <li>Attempts to revoke the claimant's representation via CCD.</li>
+     *     <li>If revocation fails and the claimant representative’s email matches
+     *         a respondent representative, forcefully removes the claimant representation.</li>
+     *     <li>Sends a notification to the claimant if the representation is successfully removed.</li>
```
btensay: I think notification is no more being sent here!

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
JustMehmet: @btensay Thanks for your feedback, Tensay.

I just wanted to clarify that we do not currently have any rollback logic implemented in our system. However, we do have logging and a retry mechanism in place. In this project, the retry is handled through an inner event call.

A similar issue could also occur in scenarios where representatives are revoked, but a new representative is not successfully assigned.

That said, if any failure occurs during an assignment, we log the failure and continue processing the remaining assignments. This is a known and accepted behaviour in the current process, and we make sure it is captured in the logs for visibility.

hsjhita (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocNotificationService.java
```diff
@@ -0,0 +1,639 @@
...
+                    e.getMessage());
+        }
+    }
+
+    private RetrieveOrgByIdResponse getOrganisationResponse(String orgId, boolean isNewOrg) {
+        ResponseEntity<RetrieveOrgByIdResponse> getOrgResponse = getOrganisationById(orgId);
+        HttpStatusCode statusCode = getOrgResponse.getStatusCode();
+        if (!HttpStatus.OK.equals(statusCode)) {
+            String orgType = isNewOrg ? "new" : "old";
+            log.error("Cannot retrieve {} org by id {} [{}] {}", orgType, orgId, statusCode, getOrgResponse.getBody());
```
Mike-Brain: Risk of PII with getOrgResponse.getBody() as it contains personal info

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/OrganisationService.java
```diff
@@ -0,0 +1,133 @@
...
+            if (!OrganisationUtils.hasOrganisationSuperuserEmail(organisationResponse)) {
+                return null;
+            }
+        } catch (FeignException e) {
+            if (e.status() != HttpStatus.NOT_FOUND.value()) {
+                log.error(WARNING_UNABLE_TO_FIND_ORGANISATION_SUPER_USER, e.getMessage());
+            }
+            return null;
+        }
+        assert organisationResponse.getBody() != null;
```
btensay, line 130: If assert is not disabled in prod, it could lead to unexpected crashes. If disabled, no null check is made. Hence, in either case, I think, an explicit null check/guard is better. The same with asserts in NocUtils and NocNotificationService.

JustMehmet: Hi Tensay,

Thank you so much for your feedback.

This should not cause a NullPointerException, because organisationResponse.getBody() is already checked for null in [OrganisationUtils.hasOrganisationSuperuserEmail](https://github.com/hmcts/et-ccd-callbacks/blob/f1376b17003af6a15536b2268a9977887c9a5156/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/OrganisationUtils.java#L457).

I added the assertion for a few reasons:

**1. To satisfy the IDE/static analyser**
Java, IntelliJ, or Sonar may not recognise that hasOrganisationSuperuserEmail(...) guarantees organisationResponse.getBody() is non-null. The assertion makes that assumption explicit from that point onwards.
**2. To document the assumption**
It makes the invariant clear locally: if the code reaches this point, getBody() is expected to be present.
**3. As a defensive check during testing/development**
If hasOrganisationSuperuserEmail(...) is changed in the future and no longer guarantees a non-null body, the assertion could help catch that during testing.

hsjhita (dismissed):

btensay (dismissed):

LahiruOshara (dismissed):

btensay (dismissed):

LahiruOshara (approved):

btensay (approved):

PR #3037, opened 2026-03-23
RET-6005, RET-6198, RET-6286

JustMehmet:
### Jira link

1. [RET-6005](https://tools.hmcts.net/jira/browse/RET-6005)
2. [RET-6198 ](https://tools.hmcts.net/jira/browse/RET-6198)

-----

### Change description

Add support for representative self-revocation from a case.

MyHMCTS/ExUI - https://xui-et-cos-pr-3037.preview.platform.hmcts.net/
CCD Admin Web - https://admin-web-et-cos-pr-3037.preview.platform.hmcts.net/
Camunda - https://camunda-et-cos-pr-3037.preview.platform.hmcts.net/
SYA Web - https://et-sya-et-cos-pr-3037.preview.platform.hmcts.net/
SYR Web - https://et-syr-et-cos-pr-3037.preview.platform.hmcts.net/
Manage Org - https://xui-mo-webapp-et-cos-pr-3037.preview.platform.hmcts.net/
Register Org - https://xui-mo-webapp-et-cos-pr-3037.preview.platform.hmcts.net/register-org/register
Approve Org - https://xui-ao-webapp-et-cos-pr-3037.preview.platform.hmcts.net/

-----

### Checklist


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationService.java
```diff
@@ -0,0 +1,194 @@
...
+                NOC_TYPE_REMOVAL
+            );
+        final String repName = userDetails.getName();
+        final List<String> removedRepsEmailAddresses = NocRespondentMapper.getRepresentativeEmails(repListToRevoke);
+        // send email to organisation admin when organisation is revoked
+        if (REMOVE_ORGANISATION.equals(caseDetails.getCaseData().getNocRemoveRepOption())) {
+            nocRemoveRepresentationEmailService.sendEmailToOrgAdmin(caseDetailsBeforeRepUpdate, orgEmailAddress,
+                    repName, EMAIL_TYPE_TO_ORG_ADMIN_REMOVED);
+        } else if (!OrganisationUtils.hasRemainingRespondentReps(caseDetails.getCaseData(),
+                organisation.getOrganisationIdentifier())) {
```
btensay: As organisation.getOrganisationIdentifier() is not covered by the ObjectUtils.isNotEmpty(organisation) check, it could potentially cause NPE.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
```diff
@@ -604,32 +645,47 @@ public void revokeAndRemoveRespondentRepresentatives(CaseDetails caseDetails,
...
         }
-
-        for (CaseUserAssignment caseUserAssignment : caseUserAssignmentsData.getCaseUserAssignments()) {
-            if (!RoleUtils.isRespondentRepresentativeRole(caseUserAssignment.getCaseRole())) {
-                continue;
+        List<CaseUserAssignment> caseUserAssignments =
+                new ArrayList<>(caseUserAssignmentsData.getCaseUserAssignments());
+        List<CaseUserAssignment> manualAssignments = new ArrayList<>();
+        for (RepresentedTypeRItem representative : representativesToRevoke) {
+            manualAssignments.addAll(RespondentRepresentativeUtils.findManualAssignments(
```
btensay, line 669: If I am reading it right, manualAssignments has to be moved into the loop or cleared at the end of each loop iteration to prevent error at line 670.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
```diff
@@ -0,0 +1,274 @@
...
+        repEmailAddress.forEach(email -> sendEmailToRemovedLegalRep(caseDetails, email));
+    }
+
+    /**
+     * Sends an email to a removed legal representative.
+     *
+     * @param caseDetails The case details containing case data and ID
+     * @param emailToSend The email address of the removed legal representative
+     */
+    public void sendEmailToRemovedLegalRep(CaseDetails caseDetails, String emailToSend) {
```
btensay, line 99:
```suggestion
    public void sendEmailToRemovedLegalRep(CaseDetails caseDetails, String emailToSend) {
            if (isNullOrEmpty(emailToSend)) {
            return;
        }
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
```diff
@@ -0,0 +1,274 @@
...
+        CaseData caseData = caseDetails.getCaseData();
+
+        // check if claimant is represented
+        RepresentedTypeC representativeClaimantType = caseData.getRepresentativeClaimantType();
+        boolean isClaimantRepresented = representativeClaimantType != null;
+
+        // get email address of claimant or claimant legal rep
+        String emailToSend = isClaimantRepresented
+            ? representativeClaimantType.getRepresentativeEmailAddress()
+            : caseData.getClaimantType().getClaimantEmailAddress();
```
btensay: I think a null guard is required if _the claimant is unrepresented and claimantType is null_

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRespondentRepresentativeService.java
JustMehmet: @btensay Thank you so much for your feedback, Tensay.

That loop has slightly complex logic because of how applyNOC currently works. As shown in the attached video, any applyNOC process automatically adds extra assignments. When removing a representative from a case, we need to:

Remove all assignments for that representative.
Remove any other representative assignments linked to that representative’s role.

That is why, in this loop, we iterate through each representative and do the following:

Find manual assignments
These are assignments created by us. In this case, the representative has assignments that also have respondents. We identify these by matching the representative’s respondent ID with the respondent, then checking the role. If the role matches the representative’s assignment role, we treat it as a manual assignment.
Find auto-generated assignments
These are assignments automatically added by the applyNOC procedure on the CCD side. Based on the warnings we raised, this has been identified as a CCD bug, and they are removing that logic on their side.
Prepare the representative for removal
Removing a representative has two main steps:
Remove the representative from the representative collection.
Revoke their assignments.

So, during the loop, we move all assignments that need to be revoked into revokedCaseUserAssignments. Once the loop finishes, we revoke those assignments and add the representatives to the representativesToRemove list, so their Organisation Policies can be removed and they can be removed from the representative collection.

[NOC BUG VIDEO](https://github.com/user-attachments/assets/b19057ae-45cb-49cc-a270-982a80faac1f)
https://github.com/user-attachments/assets/b19057ae-45cb-49cc-a270-982a80faac1f

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
```diff
@@ -0,0 +1,276 @@
...
+     */
+    public void sendEmailToUnrepresentedClaimant(CaseDetails caseDetails, String orgName) {
+        CaseData caseData = caseDetails.getCaseData();
+        if (ObjectUtils.isEmpty(caseData.getClaimantType())
+            || StringUtils.isBlank(caseData.getClaimantType().getClaimantEmailAddress())) {
+            return;
+        }
+        String emailToSend = caseData.getClaimantType().getClaimantEmailAddress();
+        String linkToCitUI = emailService.getCitizenCaseLink(caseDetails.getCaseId());
+        sendEmailToUnrepresentedParty(caseDetails, emailToSend, orgName, linkToCitUI);
```
hsjhita, line 132: Just a thought with this, if the claimant does not have access to the portal, should we be sending them an email with a link to the case? Do we need to check to see if they have access to the case?

Maybe one to check with Dan

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
```diff
@@ -0,0 +1,276 @@
...
+    public void sendRepresentationRemovedEmailToRespondents(
+        CaseDetails caseDetails,
+        List<RespondentSumTypeItem> respondents,
+        String orgName) {
+        for (RespondentSumTypeItem respondent : respondents) {
+            if (!RespondentUtils.isValidRespondent(respondent)
+                    || StringUtils.isBlank(respondent.getValue().getRespondentEmail())) {
+                continue;
+            }
+            String respondentEmailAddress = respondent.getValue().getRespondentEmail();
```
hsjhita: I think this should use `getRespondentEmailAddress` in `EmailNotificationService` as this field could be empty and we could have the email in the other field

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/OrganisationUtils.java
```diff
@@ -434,4 +458,37 @@ public static boolean hasMatchingOrganisationId(Organisation organisation,
...
+     */
+    public static boolean hasRemainingRespondentReps(CaseData caseData,
+                                                     String organisationId) {
+        if (CollectionUtils.isEmpty(caseData.getRepCollection()) || StringUtils.isBlank(organisationId)) {
+            return false;
+        }
+        return caseData.getRepCollection().stream()
+                .filter(RespondentRepresentativeUtils::isValidRepresentative)
+                .anyMatch(rep -> organisationId.equals(
+                        rep.getValue().getRespondentOrganisation().getOrganisationID()));
```
hsjhita, line 455: potential NPE on `getRespondentOrganisation`

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
JustMehmet: I agree with the feedback, but to avoid making this PR more complex, could we handle this in a separate ticket?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/OrganisationUtils.java
JustMehmet: Added check for NPE.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
JustMehmet: Implemented a new utility method, [findRespondentEmailAddress](https://github.com/hmcts/et-ccd-callbacks/blob/1068dc55a3102e8ff9839c8f790ada93224ed89d/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/utils/RespondentUtils.java#L442), in RespondentUtils.

The method checks both responseRespondentEmail and respondentEmail to identify the appropriate respondent email address. I have also reused this method in [getRespondentEmailAddress](https://github.com/hmcts/et-ccd-callbacks/blob/1068dc55a3102e8ff9839c8f790ada93224ed89d/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/EmailNotificationService.java#L109) within EmailNotificationService to keep the email lookup logic consistent.

btensay (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationService.java
```diff
@@ -0,0 +1,197 @@
...
+        CaseData caseData = caseDetails.getCaseData();
+        // get existing rep and organisation details for sending emails
+        RepresentedTypeC existingClaimantRep = caseData.getRepresentativeClaimantType();
+        if (existingClaimantRep == null) {
+            throw new IllegalStateException(String.format(EXCEPTION_REPRESENTATIVE_NOT_FOUND, caseDetails.getCaseId()));
+        }
+        CaseDetails caseDetailsBeforeRepUpdate = CaseDataUtils.cloneCaseDetails(caseDetails);
+        if (caseDetailsBeforeRepUpdate == null) {
+            return;
+        }
```
btensay: This null check might as well be done at the start of this method, as it only get null if the parameter caseDetails is null already. Otherwise, an exception while cloning does not reach it. Hence, it is a bit misleading as it seems to check if the clone returned is null.

btensay (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
hsjhita: @JustMehmet it's still a valid point - if a claimant that doesn't have portal access, why would we send them a link to a case they can't access? We should be telling them how they can access their case. If Dan's happy with this then I will accept this

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
```diff
@@ -0,0 +1,276 @@
...
+     * and processing continues for the remaining respondents.</p>
+     *
+     * @param caseDetails the case details containing the case ID and case data used to build the email
+     *                    link and common personalisation values
+     * @param respondents the respondents to notify
+     * @param partyName the party name to include in the email personalisation
+     * @throws NullPointerException if {@code caseDetails}, its case data, or {@code respondents}
+     *                              is {@code null}
+     */
+    public void sendEmailToOtherRespondents(CaseDetails caseDetails,
```
hsjhita, line 245: @JustMehmet similar point to the claimants email, we could be sending an email to someone that doesn't have portal access or they may have representation so what if the rep needs to be informed

hsjhita (commented):
See comments around emails but think there's a missing requirement
If a respondent rep was representing `respondent A` and `respondent B`, how can they remove themselves from just one respondent? Think we need an option if they just want to cease representation for just one respondent

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java
JustMehmet: Thanks for raising this. Based on [RET-6017](https://tools.hmcts.net/jira/browse/RET-6017), no separate email template is specified for other parties—whether claimants or respondents—who do not have portal access or are not represented. Where we have an email address for them, we send the standard notification email.

Separately, Pauline is creating a new ticket to cover citizens who are not registered with IDAM.

We also already notify all removed representatives through [sendEmailToListOfRemovedLegalRep](https://github.com/hmcts/et-ccd-callbacks/blob/fc2853b611b29424badb157a23ca6a99027b2873/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java#L89), and the organisation’s super user through [sendEmailToOrgAdmin](https://github.com/hmcts/et-ccd-callbacks/blob/fc2853b611b29424badb157a23ca6a99027b2873/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/service/noc/NocRemoveRepresentationEmailService.java#L60), about the representation update.

hsjhita: I think you may be mistaken - when a claimantRep is removed, we are sending emails to the respondents. But if a respondents has a rep, they are not getting any emails. So it adds to the point that not only would a potential non portal user get a link to et-syr, but if they do have representation and they have access to the portal, the rep isn't getting an email

PR #3350, opened 2026-07-27
Update_NOC_Data

JustMehmet:
### Jira link

There is no Jira link.

-----

### Change description

Updates to NoticeOfChangeFieldsTask to set representative id when it is empty, set role field of representative and set representative id field of respondent.

-----

### Checklist

### DIFF
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
@@ -62,7 +62,6 @@ public class NoticeOfChangeFieldsTask implements Runnable {
 
     @Override
     public void run() {
-        String query = buildQuery();
         String adminUserToken = adminUserService.getAdminUserToken();
         String[] caseTypeIds = caseTypeIdsString.split(",");
 
@@ -70,8 +69,10 @@ public class NoticeOfChangeFieldsTask implements Runnable {
             List<SubmitEvent> cases;
             try {
                 do {
+                    String query = buildQuery();
                     cases = ccdClient.buildAndGetElasticSearchRequest(adminUserToken, caseTypeId, query);
                     setLastCaseId(cases);
+                    log.info("Executing caseId: " + lastCaseId);
                     log.info("{} - Notice of change fields task - Retrieved {} cases", caseTypeId, cases.size());
                     updateCases(cases, caseTypeId, adminUserToken);
                 } while (CollectionUtils.isNotEmpty(cases));
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
@@ -101,6 +101,7 @@ public class NoticeOfChangeFieldsTask implements Runnable {
                     }
                 });
             }
+            executor.shutdown();
             if (!executor.awaitTermination(awaitTimeoutSeconds, TimeUnit.SECONDS)) {
                 log.warn("{} - Executor did not terminate within {}s, forcing shutdown",
                         caseTypeId, awaitTimeoutSeconds);
--- a/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
+++ b/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
@@ -11,7 +11,6 @@ import org.elasticsearch.index.query.ExistsQueryBuilder;
 import org.elasticsearch.index.query.TermsQueryBuilder;
 import org.elasticsearch.search.builder.SearchSourceBuilder;
 import org.springframework.beans.factory.annotation.Value;
-import org.springframework.scheduling.annotation.Scheduled;
 import org.springframework.stereotype.Component;
 import uk.gov.hmcts.ecm.common.client.CcdClient;
 import uk.gov.hmcts.et.common.model.ccd.CCDRequest;
@@ -22,7 +21,6 @@ import uk.gov.hmcts.et.common.model.ccd.items.RepresentedTypeRItem;
 import uk.gov.hmcts.et.common.model.ccd.items.RespondentSumTypeItem;
 import uk.gov.hmcts.ethos.replacement.docmosis.exceptions.GenericServiceException;
 import uk.gov.hmcts.ethos.replacement.docmosis.service.AdminUserService;
-import uk.gov.hmcts.ethos.replacement.docmosis.service.FeatureToggleService;
 import uk.gov.hmcts.ethos.replacement.docmosis.utils.RespondentUtils;
 import uk.gov.hmcts.ethos.replacement.docmosis.utils.noc.RespondentRepresentativeUtils;
 import uk.gov.hmcts.ethos.replacement.docmosis.utils.noc.RoleUtils;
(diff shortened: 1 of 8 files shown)


src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
```diff
@@ -54,12 +53,12 @@ public class NoticeOfChangeFieldsTask {
...
-    public void generateNoticeOfChangeFields() {
-        if (!featureToggleService.isNoticeOfChangeFieldsEnabled()) {
-            return;
-        }
+    @Override
+    public void run() {
+        generateNoticeOfChangeFields();
+    }
 
+    public void generateNoticeOfChangeFields() {
```
hsjhita:
```suggestion
    @Override
    public void run() {
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
```diff
@@ -115,33 +114,76 @@ public static String findCaseId(SubmitEvent se) {
...
+                    }
+                    if (StringUtils.isBlank(representative.getId())) {
+                        representative.setId(UUID.randomUUID().toString());
+                    }
+                    RespondentSumTypeItem respondent = RespondentRepresentativeUtils
+                         .findRespondentByRepresentative(caseData, representative);
+                    setRespondentValues(caseData, representative, respondent);
+                }
+                ccdClient.submitEventForCase(adminUserToken, caseData, caseDetails.getCaseTypeId(),
+                        caseDetails.getJurisdiction(), ccdRequest, caseDetails.getCaseId());
```
hsjhita: Could we make the execution of this conditional from a boolean? For example only submitting if there has been a data change. Say if a case already has the id etc, there's no need to submit an event.

DataQualityTask example - https://github.com/hmcts/et-ccd-callbacks/blob/master/src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/DataQualityTask.java#L215-L221

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
```diff
@@ -115,33 +110,88 @@ public static String findCaseId(SubmitEvent se) {
...
+    public static boolean setRespondentValues(CaseData caseData,
+                                    RepresentedTypeRItem representative,
+                                    RespondentSumTypeItem respondent) {
+        if (!RespondentUtils.isValidRespondent(respondent)) {
+            return false;
+        }
+        respondent.getValue().setRepresentativeId(representative.getId());
+        int respondentIndex = RespondentUtils.getRespondentIndexById(caseData, respondent.getId());
+        if (respondentIndex != NumberUtils.INTEGER_MINUS_ONE) {
+            representative.getValue().setRole(RoleUtils
```
btensay: To reduce the amount of submit events and particularly to avoid re-submitting for every matching case, how about to add a check if the role we want to set is already correct?

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
```diff
@@ -115,33 +110,88 @@ public static String findCaseId(SubmitEvent se) {
...
+    }
+
     private String buildQuery() {
         return new SearchSourceBuilder()
                 .size(maxCases)
                 .query(new BoolQueryBuilder()
                         .must(new TermsQueryBuilder("state.keyword", validStates))
                         .must(new TermsQueryBuilder("jurisdiction.keyword", EMPLOYMENT))
-                        .mustNot(new ExistsQueryBuilder("data.claimantRepresentativeOrganisationPolicy"))
+                        .must(new ExistsQueryBuilder("data.repCollection"))
```
btensay, line 230: Because the change your event makes does not affect the repCollection's existence, if you have more than 500 cases, you would not be able to navigate past 500 as there is no search_after or must_not_exist sort of filter. I suspect, every run returns the first 500 matches.

The repo already has a worked example in WaTaskCreationCronForExpiredBfActions.findCasesByCaseType (src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/wa/expiredbftask/), which sorts on reference.keyword: asc, uses the last hit's reference as the cursor, and stops at cron.maxCasesToProcess. That property already exists in application.yaml (CRON_MAX_CASES_PER_SEARCH, default 1500) and this task doesn't currently use it.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
JustMehmet: @btensay Thank you so much Tensay, corrected.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
JustMehmet: Thanks Tensay, was a very important comment. Corrected.

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
```diff
@@ -54,25 +55,26 @@ public class NoticeOfChangeFieldsTask {
     @Value("${cron.maxCasesPerSearch}")
     private int maxCases;
 
-    @Scheduled(cron = "${cron.noticeOfChange}")
-    public void generateNoticeOfChangeFields() {
-        if (!featureToggleService.isNoticeOfChangeFieldsEnabled()) {
-            return;
-        }
+    private static final String UNKNOWN_CASE_ID = "<unknown>";
+    private static final String DEFAULT_CASE_ID = "0";
```
btensay: This may risk Scottish type inheriting EW index if EW fails mid run. May be using a local **_String cursor = "0"_** per case type would be safer.

btensay (dismissed):

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
hsjhita: As mentioned on Slack, switching to Virtual Threads means this shutdown code is redundant and can be removed

```suggestion
```

src/main/java/uk/gov/hmcts/ethos/replacement/docmosis/tasks/NoticeOfChangeFieldsTask.java
```diff
@@ -109,39 +109,128 @@ private void updateCases(List<SubmitEvent> cases, String caseTypeId, String admi
...
-        return new SearchSourceBuilder()
+        SearchSourceBuilder searchSourceBuilder = new SearchSourceBuilder()
                 .size(maxCases)
                 .query(new BoolQueryBuilder()
                         .must(new TermsQueryBuilder("state.keyword", validStates))
                         .must(new TermsQueryBuilder("jurisdiction.keyword", EMPLOYMENT))
-                        .mustNot(new ExistsQueryBuilder("data.claimantRepresentativeOrganisationPolicy"))
-                ).toString();
+                        .must(new ExistsQueryBuilder("data.repCollection"))
+                ).sort("reference.keyword", SortOrder.ASC);
```
hsjhita, line 231: Minor point but should look to use source to reduce the size of data and returned by ElasticSearch and only use the fields you need

hsjhita (dismissed):
Minor comments but approved

hsjhita (dismissed):

btensay (dismissed):

btensay (dismissed):

hsjhita (dismissed):

btensay (approved):

hsjhita (approved):

PR #3371, opened 2026-08-03
Add unique index for ethos case references

jasonpaige:
## What changed

Added an ET-owned Flyway migration, `V017__CreateEthosCaseReferenceUniqueIndex.sql`, to create a unique index on `ccd.case_data` over case type and the trimmed, upper-cased `ethosCaseReference` value.

## Why

The equivalent constraint existed in the central CCD database but was not copied to ET's decentralised database. The table is managed in the SDK-owned `ccd` schema, while this service-specific index remains in ET's Flyway migrations. Adding it keeps the schema consistent and prevents case-insensitive duplicate ethos case references within a case type. The central CCD jurisdiction predicate is intentionally omitted because this database contains ET data only.

## Impact

Existing valid data is unaffected. Future inserts or updates that would create a duplicate normalised ethos case reference for the same case type will be rejected by PostgreSQL.

## Validation

- `./gradlew printVersionInit processResources`
- `git diff --check`

### DIFF
--- a/config/owasp/suppressions.xml
+++ b/config/owasp/suppressions.xml
@@ -89,6 +89,8 @@
         <packageUrl regex="true">^pkg:maven/co\.elastic\.clients/elasticsearch\-(java|rest5\-client)@9\.4\.2$</packageUrl>
         <cve>CVE-2026-56148</cve>
         <cve>CVE-2026-56149</cve>
+        <cve>CVE-2026-63144</cve>
+        <cve>CVE-2026-63263</cve>
     </suppress>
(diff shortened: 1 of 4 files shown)


banderous (dismissed):
Looks good; can can do this - exceptionally - without _concurrently_ because these tables are not yet in use

hsjhita (dismissed):

banderous (dismissed):

hsjhita (approved):

banderous (approved):

